PR #2, opened 2022-06-15
fix: next round of refactoring

peternhale:
This work represents the application of strong typing to packaging utils and supporting source. Refactoring is not complete but wanting to get this published to help with the next round of command migrations.


src/interfaces/packagingInterfacesAndType.ts
```diff
@@ -37,4 +42,31 @@ export interface IPackage2Version {
...
+  CreatedDate: string;
+  HasMetadataRemoved: boolean | null;
+  CreatedBy: string;
+};
+
+export type Package2VersionCreateRequestError = {
+  Message: string;
+};
+
+export type Package2VersionCreateEventData = {
```
peternhale, line 66: To be used for Lifecycle events during long running package version create operations.

src/interfaces/packagingSObjects.ts
```diff
@@ -0,0 +1,244 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+export namespace PackagingSObjects {
```
peternhale, line 7: SObject schema for tooling objects are not yet available via jsforce2, adding here for the time being.

src/utils/packageUtils.ts
```diff
@@ -0,0 +1,976 @@
+/*
```
peternhale, line 1: I removed the pkgutil member and pulled all functions up into their own export.

src/utils/packageUtils.ts
```diff
@@ -0,0 +1,976 @@
...
+    const versionNumber = VersionNumber.from(versionNumberString);
+    if (versionNumber.patch !== '0') {
+      throw messages.createError('errorInvalidPatchNumber', [versionNumberString]);
+    }
+  }
+}
+
+// TODO: let's get rid of this in favor of SfdcUrl.isValidUrl()
+// check that the provided url has a valid format
+export function validUrl(url: string): boolean {
```
peternhale, line 147: Covering the preferred function with old function so it makes it simpler to replace when all code has been moved over.

src/utils/packageUtils.ts
```diff
@@ -0,0 +1,976 @@
...
+      records = records.concat(result.records);
+    }
+    itemsQueried += chunkCount;
+  }
+  return records;
+}
+/**
+ *   Returns the number of items that can be included in a quoted comma-separated string (e.g., "'item1','item2'") not exceeding maxLength
+ */
+// TODO: this function cannot handle a single item that is longer than maxLength - what to do, since this could be the root cause of an infinite loop?
```
peternhale, line 475: Not sure if this function should throw when an item count of zero will be returned. I modified the calling function to throw, just so we don't get into infinite loop (very unlikely, it would take first item to have a length around 4000 to trigger the loop.

src/utils/packageUtils.ts
```diff
@@ -0,0 +1,976 @@
...
+  if (!createResult.success) {
+    throw new Error(createResult.errors.map((e) => e.message).join('\n'));
+  }
+  return createResult.id;
+}
+export function _getPackageVersionCreateRequestApi(connection: Connection): PackageVersionCreateRequestApi {
+  return new PackageVersionCreateRequestApi({ connection });
+}
+
+export async function pollForStatusWithInterval(
```
peternhale, line 825: refactored to use PollingClient. This function still needs additional refactoring

src/package/packageList.ts
```diff
@@ -0,0 +1,20 @@
...
+import { PackagingSObjects } from '../interfaces';
+const QUERY =
+  'SELECT Id, SubscriberPackageId, Name, Description, NamespacePrefix, ContainerOptions, IsOrgDependent, ConvertedFromPackageId, ' +
+  'PackageErrorUsername, CreatedById ' +
+  'FROM Package2 ' +
+  'WHERE IsDeprecated != true ' +
+  'ORDER BY NamespacePrefix, Name';
+
+export async function listPackages(connection: Connection): Promise<QueryResult<PackagingSObjects.Package2>> {
+  return await connection.tooling.query<PackagingSObjects.Package2>(QUERY);
```
shetzel, line 19: Outside of try/catch blocks, return await is redundant. There's even an [ESLint rule to detect it](https://github.com/eslint/eslint/blob/master/docs/rules/no-return-await.md), but it allows it in try/catch.

src/package/packageVersionCreateRequestApi.ts
```diff
@@ -0,0 +1,120 @@
...
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const packageUtils = require('../utils/packageUtils');
+const STATUS_ERROR = 'Error';
+const QUERY =
+  'SELECT Id, Status, Package2Id, Package2VersionId, Package2Version.SubscriberPackageVersionId, Tag, Branch, ' +
+  'CreatedDate, Package2Version.HasMetadataRemoved, CreatedById ' +
+  'FROM Package2VersionCreateRequest ' +
+  '%s' + // WHERE, if applicable
+  'ORDER BY CreatedDate';
+const ERROR_QUERY = "SELECT Message FROM Package2VersionCreateRequestError WHERE ParentRequest.Id = '%s'";
```
shetzel, line 31: (not blocking) I wonder if we should have a `queries.ts` with all the queries used in the lib with more descriptive names.  Not sure it makes sense yet but something that might reveal itself to be useful.

src/package/packageVersionCreateRequestApi.ts
```diff
@@ -0,0 +1,120 @@
...
+const STATUSES = ['Queued', 'InProgress', 'Success', 'Error'];
```
shetzel, line 32: (not blocking) This might also be useful as a library export, with a more descriptive name.

src/package/packageVersionCreateRequestApi.ts
```diff
@@ -0,0 +1,120 @@
...
+
+type PackageVersionCreateRequestApiOptions = {
+  createdlastdays?: number;
+  connection?: Connection;
+  status?: string;
+};
+
+export class PackageVersionCreateRequestApi {
```
shetzel, line 40: (not blocking) never liked this name.  I have ideas for how to rename but that can come later.

shetzel (approved):

PR #3, opened 2022-06-22
fix: add package version list

peternhale:


src/interfaces/packagingInterfacesAndType.ts
```diff
@@ -70,3 +70,39 @@ export type Package2VersionCreateEventData = {
   message?: string;
   timeRemaining?: Duration;
 };
+
+export type PackageVersionListResult = {
```
shetzel, line 74: These need exporting too right?

src/package/packageVersionList.ts
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as _ from 'lodash';
```
shetzel, line 8: lodash is another library I'd try to avoid bringing in as a dep if possible.

src/package/packageVersionList.ts
```diff
@@ -0,0 +1,133 @@
...
+  packages: string[];
+  connection: Connection;
+  verbose: boolean;
+  concise: boolean;
+  isReleased: boolean;
+}): Promise<QueryResult<PackageVersionListResult>> {
+  return options.connection.tooling.query<PackageVersionListResult>(_constructQuery(options));
+}
+
+export function _constructQuery(options: {
```
shetzel, line 50: Why do some functions begin with an underscore?  Were those "private"?

src/package/packageVersionList.ts
```diff
@@ -0,0 +1,133 @@
...
+    lastModLastDays = _getLastDays('modifiedlastdays', lastModLastDays);
+    where.push(`LastModifiedDate = LAST_N_DAYS:${lastModLastDays}`);
+  }
+
+  // exclude deleted
+  where.push('IsDeprecated = false');
+  return where;
+}
+
+export function _getLastDays(paramName: string, lastDays: number) {
```
shetzel, line 123: nit: missing return typing

src/package/packageVersionList.ts
```diff
@@ -0,0 +1,133 @@
...
+  project: SfProject;
+  orderBy: string;
+  modifiedLastDays: number;
+  createdLastDays: number;
+  packages: string[];
+  connection: Connection;
+  verbose: boolean;
+  concise: boolean;
+  isReleased: boolean;
+}): string {
```
shetzel, line 60: Maybe we should create `PackageQueryOptions` type?

shetzel (approved):

src/interfaces/packagingInterfacesAndType.ts
peternhale: @shetzel not following, the index.ts is reexporting everything in this file and so-on up to src/exported.ts.

src/package/packageVersionList.ts
peternhale: Ok, I will work on that for my next PR.

PR #10, opened 2022-07-28
Wr/package1 version create

WillieRuemmele:
@W-10833096@

adds a method to create a 1gp package upload request for `package1:version:create`
refactors tests
allows `.test.ts` extension to be run with `yarn test`


peternhale (changes requested):

peternhale (approved):

.sfdevrc.json
```diff
@@ -1,5 +1,5 @@
 {
   "test": {
-    "testsPath": "test/**/*Test.ts"
+    "testsPath": "test/**/*est.ts"
```
shetzel, line 3: it's better to just have 1 pattern.  Let's only support *.test.ts.  That is pretty much universal in our plugins.

src/interfaces/packagingInterfacesAndType.ts
```diff
@@ -223,3 +223,16 @@ export type PackageVersionReportResult = Partial<PackagingSObjects.Package2Versi
...
+  | 'MetadataPackageId'
+  | 'VersionName'
+  | 'Description'
+  | 'MajorVersion'
+  | 'MinorVersion'
+  | 'IsReleaseVersion'
+  | 'ReleaseNotesUrl'
+  | 'PostInstallUrl'
+  | 'Password'
+>;
```
shetzel, line 238: Are all of these fields required to create a package1 version?  I suspect not.  I think you'll have to do an intersection of the required fields and optional fields.  something like:
`Partial<Pick<PackagingSObjects.PackageUploadRequest, {optional_fields}>> & Pick<PackagingSObjects.PackageUploadRequest, {required_fields}>`

src/package1/package1VersionCreate.ts
```diff
@@ -0,0 +1,62 @@
...
+        throw messages.createError('package1VersionCreateCommandUploadFailureDefault');
+      }
+    }
+  }
+};
+
+export async function package1VersionCreate(
+  connection: Connection,
+  options: Package1VersionCreateRequest,
+  pollingOptions = { frequency: Duration.seconds(5), timeout: Duration.seconds(0) }
```
shetzel, line 46: It'll be a better API if it supports both `number` and `Duration` as polling options.  We don't have to change this now but I'd want the GA version to handle that.  I have a version of it in my branch if you're interested.

src/package1/package1VersionCreate.ts
```diff
@@ -0,0 +1,62 @@
...
+    const pollingClient = await PollingClient.create({
+      poll: () => packageUploadRequestStatus(createRequest.id, connection, timeout, pollingOptions.frequency.seconds),
+      ...pollingOptions,
+    });
+    return pollingClient.subscribe<PackagingSObjects.PackageUploadRequest>();
+  } else {
+    // jsforce templates weren't working when setting the type to PackageUploadRequest, so we have to cast `as unknown as PackagingSObjects.PackageUploadRequest`
+    return (await connection.tooling
+      .sobject('PackageUploadRequest')
+      .retrieve(createRequest.id)) as unknown as PackagingSObjects.PackageUploadRequest;
```
shetzel, line 60: Yeah, I saw this too.  We should try to fix this in jsforce because this is not great.

src/package1/package1VersionCreate.ts
```diff
@@ -0,0 +1,62 @@
...
+  pollingOptions = { frequency: Duration.seconds(5), timeout: Duration.seconds(0) }
+): Promise<PackagingSObjects.PackageUploadRequest> {
+  const createRequest = await connection.tooling.sobject('PackageUploadRequest').create(options);
+  if (pollingOptions.timeout.seconds) {
+    const timeout = pollingOptions.timeout.seconds;
+    const pollingClient = await PollingClient.create({
+      poll: () => packageUploadRequestStatus(createRequest.id, connection, timeout, pollingOptions.frequency.seconds),
+      ...pollingOptions,
+    });
+    return pollingClient.subscribe<PackagingSObjects.PackageUploadRequest>();
```
shetzel, line 55: What happens when the pollingClient times out?  It'll throw an error but not return the last known status of the request, which can be pretty useful.  We need some common patterns to follow for polling so it's handled consistently, including emitting events, polling options, and error handling.

test/package1/package1VersionCreate.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  let sobjectStub: sinon.SinonStub;
+
+  beforeEach(async () => {
+    stubContext($$);
+    await $$.stubAuths(testOrg);
+    conn = await testOrg.getConnection();
+    sobjectStub = $$.SANDBOX.stub(conn.tooling, 'sobject')
+      .onFirstCall()
+      .returns({
+        // @ts-ignore - to avoid stubbing every property of sobject
```
shetzel, line 55: jsyk, ts-sinon handles this nicely.  I think it's `stubInterface` or something like that.

test/package1/package1VersionList.test.ts
```diff
@@ -4,101 +4,93 @@
...
+    MetadataPackageVersionId: '04t46000001ZfaXXXY',
+    Name: 'Summer 22',
+    ReleaseState: 'Beta',
+    Version: '1.0.4',
+  },
+];
+
 describe('Package1 Version List', () => {
-  const $$ = testSetup();
+  // const $$ = testSetup();
```
shetzel, line 55: should this commented code be removed?

shetzel (changes requested):

shetzel (approved):

PR #11, opened 2022-08-02
Wr/package uninstall

WillieRuemmele:
@W-10832862@

adds `uninstallPackage` method to uninstall a package, will optionally wait for uninstall to complete, by default won't poll
UTs

### DIFF
--- a/src/package/packageUninstall.ts
+++ b/src/package/packageUninstall.ts
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as os from 'os';
+import { Connection, Lifecycle, Messages, PollingClient, SfError, StatusResult } from '@salesforce/core';
+import { Duration } from '@salesforce/kit';
+import { PackagingSObjects } from '../interfaces';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/packaging', 'messages');
+
+type UninstallResult = PackagingSObjects.SubscriberPackageVersionUninstallRequest;
+
+async function poll(id: string, conn: Connection): Promise<StatusResult> {
+  const uninstallRequest = await conn.tooling.sobject('SubscriberPackageVersionUninstallRequest').retrieve(id);
+
+  switch (uninstallRequest.Status) {
+    case 'Success': {
+      return { completed: true, payload: uninstallRequest };
+    }
+    case 'InProgress':
+    case 'Queued': {
+      Lifecycle.getInstance().emit('packageUninstall', {
+        ...uninstallRequest,
+      });
+      return { completed: false, payload: uninstallRequest };
+    }
+    default: {
+      const err = messages.getMessage('defaultErrorMessage', [id, uninstallRequest.Id]);
+      const errorQueryResult = await conn.tooling.query(
+        `"SELECT Message FROM PackageVersionUninstallRequestError WHERE ParentRequest.Id = '${id}' ORDER BY Message"`
+      );
+
+      const errors = [];
+      if (errorQueryResult.records) {
+        errors.push('\n=== Errors\n');
+        errorQueryResult?.records?.forEach((record) => {
+          errors.push(`(${errors.length + 1}) ${record.Message}${os.EOL}`);
+        });
+      }
+
+      throw new SfError(`${err}${errors.join(os.EOL)}`, 'UNINSTALL_ERROR', [
+        messages.getMessage('uninstallErrorAction'),
+      ]);
+    }
+  }
+}
+
+export async function uninstallPackage(
+  id: string,
+  conn: Connection,
+  wait: Duration = Duration.seconds(0)
+): Promise<UninstallResult> {
+  const uninstallRequest = await conn.tooling.sobject('SubscriberPackageVersionUninstallRequest').create({
+    SubscriberPackageVersionId: id,
+  });
+
+  if (wait.seconds === 0) {
+    return (await conn.tooling
+      .sobject('SubscriberPackageVersionUninstallRequest')
+      .retrieve(uninstallRequest.id)) as UninstallResult;
+  } else {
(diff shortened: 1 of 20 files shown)


src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+    default: {
+      const err = messages.getMessage('defaultErrorMessage', [id, uninstallRequest.Id]);
+      const errorQueryResult = await conn.tooling.query(
+        `"SELECT Message FROM PackageVersionUninstallRequestError WHERE ParentRequest.Id = '${id}' ORDER BY Message"`
+      );
+
+      const errors = [];
+      if (errorQueryResult.records.length) {
+        errors.push('\n=== Errors\n');
+        errorQueryResult?.records?.forEach((record) => {
```
peternhale, line 40: @WillieRuemmele are the safe navigation operators necessary here?

src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  if (wait.seconds === 0) {
+    return (await conn.tooling
+      .sobject('SubscriberPackageVersionUninstallRequest')
+      .retrieve(uninstallRequest.id)) as UninstallResult;
+  } else {
+    const pollingClient = await PollingClient.create({
+      poll: () => poll(uninstallRequest.id, conn),
+      frequency: Duration.seconds(5),
```
peternhale, line 68: @WillieRuemmele is an uninstall runtime pretty short, if so then 5 second interval would be ok, but if takes a long time then maybe this value is too short.

peternhale:
@WillieRuemmele could you add the uninstall API call to this [nut test](https://github.com/forcedotcom/packaging/blob/135d4206ea130176053dd574c6b05ca9d78191bc/test/package/packageTest.nut.ts). There is an uninstall step that currently calls out to the sfdx command.

peternhale (changes requested):

src/package/packageUninstall.ts
WillieRuemmele: they were always pretty quick for me, it's also what toolbelt was using

mshanemc (commented):
I have some meta-comments about how this is organized, not specific to this PR. Hopefully it's not too late to influence without making a mess.

1. For well-known types for the Tooling API, why not put them in jsforce instead of this library?
ex: https://github.com/jsforce/jsforce/blob/2.0/src/api/tooling.ts

2.  If that happened, you could do things the jsforce2 way ex:
```ts
const uninstallRequest = await conn.tooling
    .sobject<SubscriberPackageVersionUninstallRequest>('SubscriberPackageVersionUninstallRequest')
    .retrieve(id);
```
3. Let's avoid using enums, prefer string union types, so not:
```ts
export enum Package2VersionStatus {
    queued = 'Queued',
    inProgress = 'InProgress',
    ...
  }
```
instead
```
export type Package2VersionStatus = 'Queued' | 'InProgress' | ...
```

src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as os from 'os';
+import { Connection, Lifecycle, Messages, PollingClient, SfError, StatusResult } from '@salesforce/core';
+import { Duration } from '@salesforce/kit';
+import { PackagingSObjects } from '../interfaces';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/packaging', 'messages');
```
shetzel, line 13: We don't necessarily have to change this now, but having 1 big messages markdown file is not what we should do.

src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type UninstallResult = PackagingSObjects.SubscriberPackageVersionUninstallRequest;
+
+async function poll(id: string, conn: Connection): Promise<StatusResult> {
+  const uninstallRequest = await conn.tooling.sobject('SubscriberPackageVersionUninstallRequest').retrieve(id);
+
+  switch (uninstallRequest.Status) {
+    case 'Success': {
```
shetzel, line 21: Is this the actual casing returned by the server?  I'm guessing it is but that's not consistent with the install API, which returns status in all caps.  🙄

src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+      if (errorQueryResult.records.length) {
+        errors.push('\n=== Errors\n');
+        errorQueryResult.records.forEach((record) => {
+          errors.push(`(${errors.length}) ${record.Message}${os.EOL}`);
+        });
+      }
+
+      throw new SfError(`${err}${errors.join(os.EOL)}`, 'UNINSTALL_ERROR', [
+        messages.getMessage('uninstallErrorAction'),
+      ]);
```
shetzel, line 47: You're right, this is a common function between install, install report, and uninstall.  We'll refactor later.

src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+}
+
+export async function uninstallPackage(
+  id: string,
+  conn: Connection,
+  wait: Duration = Duration.seconds(0)
```
shetzel, line 55: Something to make consistent at some point is allowing both `number` and `Duration` for polling.  It makes for a friendlier API.

src/package/packageUninstall.ts
```diff
@@ -0,0 +1,73 @@
...
+    return (await conn.tooling
+      .sobject('SubscriberPackageVersionUninstallRequest')
+      .retrieve(uninstallRequest.id)) as UninstallResult;
+  } else {
+    const pollingClient = await PollingClient.create({
+      poll: () => poll(uninstallRequest.id, conn),
+      frequency: Duration.seconds(5),
+      timeout: wait,
+    });
+    return pollingClient.subscribe();
```
shetzel, line 71: Without try/catch around this the polling client errors are thrown directly.  This is another place where we should be consistent.

shetzel (approved):

src/package/packageUninstall.ts
WillieRuemmele: it is :(

peternhale (approved):

PR #14, opened 2022-08-08
Wr/package installed list

WillieRuemmele:
@W-11542978@

adds query for `package:installed:list` command
changes NUT to use lib instead of command

### DIFF
--- a/src/interfaces/packagingSObjects.ts
+++ b/src/interfaces/packagingSObjects.ts
@@ -268,6 +268,15 @@ export namespace PackagingSObjects {
     MinPackageVersionId: string;
   };
 
+  export type InstalledPackage = {
+    Id: string;
+    SubscriberPackageId: string;
+    SubscriberPackageVersionId: string;
+    MinPackageVersionId: string;
+    SubscriberPackage?: SubscriberPackage;
+    SubscriberPackageVersion?: Omit<MetadataPackageVersion, 'MetadataPackageId' | 'ReleaseState' | 'IsDeprecated'>;
+  };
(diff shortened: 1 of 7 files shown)


src/interfaces/packagingSObjects.ts
```diff
@@ -321,6 +321,15 @@ export namespace PackagingSObjects {
     MinPackageVersionId: string;
   };
 
+  export type InstalledPackage = {
```
peternhale, line 324: @WillieRuemmele this structure does not look like the [public docs](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_installedpackage.htm).

The query that is using this looks like it is get data from relationship. I think this type belongs in `[packagingInterfacesAndType.ts](https://github.com/forcedotcom/packaging/blob/wr/packageInstalledList/src/interfaces/packagingInterfacesAndType.ts)`

peternhale (changes requested):

peternhale (approved):

mshanemc:
[fyi: fixed some conflicts and bumped ts-node to solve build errors]

QA notes:
✅ -h and --help work
❌  the sfdx command has some examples and this one doesn't. maybe use `examples` instead of `help` prop

👎🏻 one long-named package makes the whole table awful.   Any chance new oclif table can be smarter with truncate
Screen Shot 2022-08-09 at 4 28 15 PM

🏆 works without a project!  That's such a big improvement!!

✅ json matches sfdx, including the "null" behavior

PR #16, opened 2022-08-11
fix: package version. now stored in packdir in sfdx project

peternhale:
@W-11489082@

### DIFF
--- a/src/package/packageVersion.ts
+++ b/src/package/packageVersion.ts
@@ -245,8 +245,7 @@ export class PackageVersion {
     wait: Duration,
     pollInterval: Duration
   ): Promise<Duration> {
-    let w = wait;
-    // If we are polling check to seeif the package is Org-Dependent, if so, update the poll time
+    // If we are polling check to see if the package is Org-Dependent, if so, update the poll time
     if (wait.milliseconds > 0) {
       const query = `SELECT IsOrgDependent FROM Package2 WHERE Id = '${resolvedPackageId}'`;
       try {
@@ -254,13 +253,13 @@ export class PackageVersion {
           tooling: true,
         });
         if (pkgQueryResult.IsOrgDependent) {
-          w = Duration.seconds((60 / pollInterval.seconds) * wait.seconds);
+          return Duration.seconds((60 / pollInterval.seconds) * wait.seconds);
         }
       } catch {
         // do nothing
       }
     }
-    return w;
+    return wait;
   }
 
   private async updateProjectWithPackageVersion(
--- a/src/package/packageVersion.ts
+++ b/src/package/packageVersion.ts
@@ -296,12 +296,7 @@ export class PackageVersion {
       getConfigPackageDirectory(withProject.getPackageDirectories(), 'id', pkg.Id) ?? ({} as NamedPackageDir);
     pkgDir.versionNumber = packageVersionVersionString;
     pkgDir.versionDescription = packageVersion.Description;
-    const packageDirs = withProject.getPackageDirectories().map((pd) => {
-      if (pkgDir['id'] === pd['id']) {
-        return pkgDir;
-      }
-      return pd;
-    });
+    const packageDirs = withProject.getPackageDirectories().map((pd) => pkgDir.id === pd.id ? pkgDir : pd)
     withProject.getSfProjectJson().set('packageDirectories', packageDirs);
   }
 }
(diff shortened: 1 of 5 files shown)


src/package/packageVersion.ts
```diff
@@ -217,4 +231,78 @@ export class PackageVersion {
...
+          tooling: true,
+        });
+        if (pkgQueryResult.IsOrgDependent) {
+          w = Duration.seconds((60 / pollInterval.seconds) * wait.seconds);
+        }
+      } catch {
+        // do nothing
+      }
+    }
+    return w;
```
mshanemc, line 263: the let-free version
```suggestion
    // If we are polling check to see if the package is Org-Dependent, if so, update the poll time
    if (wait.milliseconds > 0) {
      const query = `SELECT IsOrgDependent FROM Package2 WHERE Id = '${resolvedPackageId}'`;
      try {
        const pkgQueryResult = await this.connection.singleRecordQuery<PackagingSObjects.Package2>(query, {
          tooling: true,
        });
        if (pkgQueryResult.IsOrgDependent) {
          return Duration.seconds((60 / pollInterval.seconds) * wait.seconds);
        }
      } catch {
        // do nothing
      }
    }
    return wait;
```

src/package/packageVersion.ts
```diff
@@ -217,4 +231,78 @@ export class PackageVersion {
...
     return updateResult;
   }
+
+  /**
+   * Increase the wait time for a package version that is org dependent.
+   *
+   * @param resolvedPackageId
+   * @param pollInterval
+   * @param wait
+   * @private
```
mshanemc, line 241: these don't add anything beyond what TS is gonna provide.

src/package/packageVersion.ts
```diff
@@ -217,4 +231,78 @@ export class PackageVersion {
...
+    }
+  }
+
+  private async generatePackageDirectory(
+    packageVersion: PackagingSObjects.Package2Version,
+    withProject: SfProject,
+    packageVersionVersionString: string
+  ) {
+    const pkg = await (await Package.create({ connection: this.connection })).getPackage(packageVersion.Package2Id);
+    const pkgDir =
```
mshanemc, line 296: putting some appropriate overloads on the utils for this class might help TS know the return type (should match the T of the first argument or be undefined).

I'm not a fan of `{} as NamedPackageDir` (it's just not).

If `getConfigPackageDirectory` returns an undefined, this method should handle that.

src/utils/packageUtils.ts
```diff
@@ -942,9 +942,8 @@ async function _generatePackageAliasEntry(
   let packageName;
   if (!aliasForPackageId || aliasForPackageId.length === 0) {
     const query = `SELECT Name FROM Package2 WHERE Id = '${packageId}'`;
-    packageName = await connection.tooling
-      .query<PackagingSObjects.Package2>(query)
-      .then((pkgQueryResult) => pkgQueryResult.records[0]?.Name);
+    const package2 = await connection.singleRecordQuery<PackagingSObjects.Package2>(query, { tooling: true });
```
mshanemc, line 945: +1

src/package/packageVersion.ts
```diff
@@ -217,4 +231,78 @@ export class PackageVersion {
...
+      getConfigPackageDirectory(withProject.getPackageDirectories(), 'id', pkg.Id) ?? ({} as NamedPackageDir);
+    pkgDir.versionNumber = packageVersionVersionString;
+    pkgDir.versionDescription = packageVersion.Description;
+    const packageDirs = withProject.getPackageDirectories().map((pd) => {
+      if (pkgDir['id'] === pd['id']) {
+        return pkgDir;
+      }
+      return pd;
+    });
```
mshanemc, line 305: if the types are correct, there should be an id, right?
```suggestion
    const packageDirs = withProject.getPackageDirectories().map((pd) => pkgDir.id === pd.id ? pkgDir : pd)
```

mshanemc (approved):
no objections, just ideas

src/package/packageVersion.ts
peternhale: @mshanemc Actually `id` is not a property of a package dir according to the formal schema for sfdx-project.json. This property is injected during package version create. I will check with packaging team as to why it is needed and should it be part of the formal schema.

WillieRuemmele (approved):

PR #20, opened 2022-08-18
Wr/package version promote

WillieRuemmele:
@W-10833194@

add Package.promote method

### DIFF
--- a/src/package/package.ts
+++ b/src/package/package.ts
@@ -6,7 +6,7 @@
  */
 import { Messages, sfdc } from '@salesforce/core';
 import { AsyncCreatable, Duration } from '@salesforce/kit';
-import { QueryResult } from 'jsforce';
+import { QueryResult, SaveResult } from 'jsforce';
 import { Optional } from '@salesforce/ts-types';
 import { IPackage, PackageOptions, PackagingSObjects } from '../interfaces';
 import {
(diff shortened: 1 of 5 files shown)


src/package/package.ts
```diff
@@ -6,7 +6,7 @@
  */
 import { Messages, sfdc } from '@salesforce/core';
 import { AsyncCreatable, Duration } from '@salesforce/kit';
-import { QueryResult } from 'jsforce';
+import { QueryResult, SaveResult } from 'jsforce';
```
peternhale, line 9: Point for discussion. I aliased jsforce SaveResult with `PackageSaveResult` to avoid having to import jsforce on the client side. Good idea or bad? @WillieRuemmele @shetzel

peternhale:
@WillieRuemmele is there a spot in current NUTS to promote a package?

peternhale (approved):
maybe add promote to NUTs

src/package/package.ts
```diff
@@ -66,6 +66,10 @@ export class Package extends AsyncCreatable<PackageOptions> implements IPackage
     return Promise.resolve(undefined);
   }
 
+  public async promote(id: string): Promise<SaveResult> {
```
peternhale, line 69: @WillieRuemmele Seems that this should be in the PackageVersion class.

peternhale (changes requested):

peternhale (approved):

PR #27, opened 2022-08-25
fix: add Package.update method, NUT

WillieRuemmele:
@W-10832873@
adds `Package.update` method
adds NUT

### DIFF
--- a/src/package/package.ts
+++ b/src/package/package.ts
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { Messages, sfdc } from '@salesforce/core';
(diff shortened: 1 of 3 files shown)


src/package/package.ts
```diff
@@ -95,8 +97,15 @@ export class Package extends AsyncCreatable<PackageOptions> implements IPackage
     return Promise.resolve(undefined);
   }
 
-  public update(): Promise<void> {
-    return Promise.resolve(undefined);
+  public async update(options: PackageUpdateOptions): Promise<PackageSaveResult> {
+    // filter out any undefined values and their keys
+    Object.keys(options).forEach((key) => options[key] === undefined && delete options[key]);
```
shetzel, line 102: jsyk, this mutates the `options` object passed in, which some would argue is an anti-pattern.  It's probably fine but an alternative would be to use the filter function and assign to a new const that is used in the tooling update call.

shetzel (approved):

PR #30, opened 2022-08-30
fix: add Package.uninstallReport method, NUT

WillieRuemmele:
@W-11670106@

adds `Package.uninstallReport` to get the status of an uninstall request
adds a NUT
cleans up typing for `tooling.retrieve` calls


test/package/packageTest.nut.ts
```diff
@@ -369,13 +367,6 @@ describe('Integration tests for @salesforce/packaging library', function () {
       ]);
     });
 
-    it('gets an error trying to uninstall again (and waiting for the result)', () => {
```
WillieRuemmele, line 372: this command doesn't have a `--wait` flag, and so it was always getting `exitCode:1`

peternhale (approved):

PR #33, opened 2022-09-06
fix: add support for ancestry

peternhale:
@W-10833165@


src/package/packageAncestry.ts
```diff
@@ -0,0 +1,517 @@
...
+    }
+    this.display(this.options ? this.options['logger'] : undefined);
+  }
+
+  private createLabel(node: AncestryTreeProducer): string {
+    const subscriberId =
+      this.verbose && node?.options?.node?.SubscriberPackageVersionId
+        ? ` (${node.options.node.SubscriberPackageVersionId})`
+        : '';
+    const label = node?.label ? `${node.label}${subscriberId}` : 'root';
```
WillieRuemmele, line 343: NIT: could be inline `return node?.label ? ...`

src/package/packageAncestry.ts
```diff
@@ -0,0 +1,517 @@
...
+        producer.children.forEach((child) => dotLines.push(AncestryDotProducer.buildDotEdge(producer, child)));
+      }
+      producers.push(...producer.children);
+    }
+    return `strict graph G {\n${dotLines.join('\n')}}`;
+  }
+}
+
+export class PackageAncestryNode extends AsyncCreatable<PackageAncestryNodeOptions> implements Attributes {
+  private readonly _version: VersionNumber;
```
WillieRuemmele, line 446: do we want to keep all of the underscores on private attributes?

test/package/packageTest.nut.ts
```diff
@@ -25,6 +26,13 @@ import { deletePackage } from '../../src/package';
 import { PackageVersion } from '../../src/package';
 import { Package } from '../../src/package';
 import { PackagingSObjects } from '../../src/interfaces';
+import { VersionNumber } from '../../src/utils/versionNumber';
+import { AncestryJsonProducer, PackageAncestry } from '../../src/package/packageAncestry';
+import { AncestryTreeProducer } from '../../src/package/packageAncestry';
+import {
+  AncestryRepresentationProducer,
```
WillieRuemmele, line 33: NIT: should be able to import all of these from `../../src/package` or `/src/interfaces` and combine a few above

WillieRuemmele (changes requested):
empty `package-ancestry.md` file

src/package/packageAncestry.ts
peternhale: What do you suggest as an alternative?

WillieRuemmele: I haven't been prefixing private methods or attributes with anything, just letting the `private` scope take care of it. 🤷‍♂️ either way

WillieRuemmele:
@peternhale I'm still seeing an empty `messages/package-ancestry.md` file, did you want to keep that for something?

WillieRuemmele (approved):

PR #36, opened 2022-09-12
Wr/standardize event names

WillieRuemmele:
@W-11740858@

standardizes lifecycle event names to `<Package | PackageVersion>/<action>-<status>`
like
`Package/create-error`
`PackageVersion/update-in-progress`


src/package/packageConvert.ts
```diff
@@ -122,3 +139,102 @@ async function createRequestObject(
...
+            if (results[0].Error.length > 1) {
+              results[0].Error.forEach((error) => {
+                // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+                errors.push(`(${errors.length + 1}) ${error}`);
+              });
+              errors.unshift(messages.getMessage('versionCreateFailedWithMultipleErrors'));
+            }
+            status = errors.length !== 0 ? errors.join('\n') : results[0].Error.join('\n');
+          }
+          await Lifecycle.getInstance().emit('Package/convert-error', { id, status });
```
peternhale, line 202: @WillieRuemmele would create exported constants, or enum or, string literal type so consumer can use those instead of having to known the exact string.

peternhale (changes requested):

peternhale (approved):

PR #52, opened 2022-09-19
fix: consolidate code that determines package type

peternhale:
@W-11778852@
@W-11784677@ [pkg] Remove references to functions applyErrorAction and massageErrorMessage from plugin-packaging

### DIFF
--- a/src/utils/packageUtils.ts
+++ b/src/utils/packageUtils.ts
@@ -138,6 +138,31 @@ export function applyErrorAction(err: Error): Error {
 
   return err;
 }
+
+export function massageErrorMessage(err: Error): Error {
+  if (err.name === 'INVALID_OR_NULL_FOR_RESTRICTED_PICKLIST') {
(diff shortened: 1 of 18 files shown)


src/utils/packageUtils.ts
```diff
@@ -167,20 +192,42 @@ export function escapeInstallationKey(key?: string): Nullable<string> {
 }
 
 /**
- * Given 0Ho the package type type (Managed, Unlocked, Locked(deprecated?))
+ * Fetch the PackageType for a given package version ID
```
shetzel, line 195: Do you think this could fit as a static method on the `Package` class or does the `04t` ID cause  it to not make sense?

shetzel (approved):

src/utils/packageUtils.ts
peternhale: I would like to see this go away.

We should expose getter on package2 for this field

We should expose getter on package version for package type

Not sure if that work is out of scope.

PR #61, opened 2022-09-28
package1 refactor

WillieRuemmele:
@W-11825243@

refactors `Package1` to `Package1Version` because it only deals with 1gp package versions
moves all methods into a class
adds some UTs for invalid IDs
adds ID validation


src/package1/package1Version.ts
```diff
@@ -0,0 +1,146 @@
...
+   * Will create a PackageUploadRequest object based on the options provided, will poll for completion if pollingOptions are provided
+   *
+   * @param options: Package1VersionCreateRequest options for the new PackageUploadRequest to be created with
+   * @param pollingOptions: options to set frequency, and duration of polling. Default to not poll
+   */
+  public async create(
+    options: Package1VersionCreateRequest,
+    pollingOptions = { frequency: Duration.seconds(5), timeout: Duration.seconds(0) }
+  ): Promise<PackagingSObjects.PackageUploadRequest> {
+    const createRequest = await this.connection.tooling.sobject('PackageUploadRequest').create(options);
```
peternhale, line 108: The create is blindly using the options as crafted by the caller. I understand that the plugin command is vetting the inputs, but if used by an API consumer is the exception thrown by the tooling create sufficient to properly guide the caller to the mistake in input values?

src/package1/package1Version.ts
```diff
@@ -0,0 +1,146 @@
...
+   * @param connection Connection to the org
+   * @param id: id of the MetadataPackageVersion sObject (starts with 04t)
+   */
+  public static async display(connection: Connection, id: string): Promise<Package1Display[]> {
+    if (!id.startsWith('04t')) {
+      throw messages.createError('invalid04tId', [id]);
+    }
+    const query = `SELECT Id,MetadataPackageId,Name,ReleaseState,MajorVersion,MinorVersion,PatchVersion,BuildNumber FROM MetadataPackageVersion WHERE id = '${id}'`;
+    const results = (await connection.tooling.query<PackagingSObjects.MetadataPackageVersion>(query)).records;
+    return results.map((result) => ({
```
peternhale, line 59: Why not return the entire object. This function's use is no longer tied to just one CLI command.

src/package1/package1Version.ts
```diff
@@ -0,0 +1,146 @@
...
+      .retrieve(id)) as unknown as PackagingSObjects.PackageUploadRequest;
+  }
+
+  /**
+   * Executes server-side logic for the package1:display command
+   *
+   * @param connection Connection to the org
+   * @param id: id of the MetadataPackageVersion sObject (starts with 04t)
+   */
+  public static async display(connection: Connection, id: string): Promise<Package1Display[]> {
```
peternhale, line 53: Consider a different name for this function, maybe `getPackageVersion`

src/package1/package1Version.ts
```diff
@@ -0,0 +1,146 @@
...
+    }));
+  }
+
+  /**
+   * Will create a PackageUploadRequest object based on the options provided, will poll for completion if pollingOptions are provided
+   *
+   * @param options: Package1VersionCreateRequest options for the new PackageUploadRequest to be created with
+   * @param pollingOptions: options to set frequency, and duration of polling. Default to not poll
+   */
+  public async create(
```
peternhale, line 104: Should probably be static. This function is creating a new 04t, which is what package1Version is supposed to represent.

src/package1/package1Version.ts
```diff
@@ -0,0 +1,146 @@
...
+      return pollingClient.subscribe<PackagingSObjects.PackageUploadRequest>();
+    } else {
+      // jsforce templates weren't working when setting the type to PackageUploadRequest, so we have to cast `as unknown as PackagingSObjects.PackageUploadRequest`
+      return (await this.connection.tooling
+        .sobject('PackageUploadRequest')
+        .retrieve(createRequest.id)) as unknown as PackagingSObjects.PackageUploadRequest;
+    }
+  }
+
+  private async packageUploadPolling(id: string, timeout: number, frequency: number): Promise<StatusResult> {
```
peternhale, line 124: Should be static

src/package1/package1Version.ts
```diff
@@ -0,0 +1,146 @@
...
+      // we have to check that it is present, and starts with 033
+      // otherwise, undefined doesn't start with 033 and will trigger this error, when it shouldn't
+      throw messages.createError('invalid033Id', [id]);
+    }
+    const query = `SELECT Id,MetadataPackageId,Name,ReleaseState,MajorVersion,MinorVersion,PatchVersion,BuildNumber FROM MetadataPackageVersion ${
+      id ? `WHERE MetadataPackageId = '${id}'` : ''
+    } ORDER BY MetadataPackageId, MajorVersion, MinorVersion, PatchVersion, BuildNumber`;
+
+    const queryResult = await connection.tooling.query<PackagingSObjects.MetadataPackageVersion>(query);
+    return queryResult.records?.map((record) => ({
```
peternhale, line 88: Return all of a MetadataPackageVersion. Caller will decide which fields to use.

src/package1/package1Version.ts
peternhale: This function should not static. Caller knows the ID so
```
const p1v = new Package1Version(id, connection);
const p1vData = p1v.getPackageVersion();
...
```

peternhale (changes requested):

src/package1/package1Version.ts
```diff
@@ -0,0 +1,142 @@
...
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/packaging', 'package1Version');
+
+/**
+ * Package1Version class - Class to be used with 1st generation package versions
+ *
+ * implementation examples can be seen here: https://github.com/salesforcecli/plugin-packaging/tree/main/src/commands/force/package1/
+ */
+export class Package1Version implements IPackageVersion1GP {
+  public constructor(private connection: Connection) {}
```
peternhale, line 27: Please add id to c'tor.

src/package1/package1Version.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+  }
+
+  /**
+   * Executes server-side logic for the package1:display command
+   *
+   * @param id: id of the MetadataPackageVersion sObject (starts with 04t)
+   */
+  public async getPackageVersion(id: string): Promise<MetadataPackageVersion[]> {
```
peternhale, line 135: Instance method can use id field, so id can be removed from this getter

peternhale (changes requested):
Looks good @WillieRuemmele, just a minor tweak to consider.

src/package1/package1Version.ts
```diff
@@ -0,0 +1,147 @@
...
+ * implementation examples can be seen here: https://github.com/salesforcecli/plugin-packaging/tree/main/src/commands/force/package1/
+ */
+export class Package1Version implements IPackageVersion1GP {
+  /**
+   * Package1Version Constructor - Class to be used with 1st generation package versions
+   *
+   * @param connection: Connection to the org
+   * @param id: 04t ID of the package version
+   */
+  public constructor(private connection: Connection, private id: string) {
```
shetzel, line 33: Can this also be an alias or only ID?

src/package1/package1Version.ts
```diff
@@ -0,0 +1,147 @@
...
+          ]);
+        } else {
+          throw messages.createError('package1VersionCreateCommandUploadFailureDefault');
+        }
+      }
+    }
+  }
+
+  /**
+   * Executes server-side logic for the package1:display command
```
shetzel, line 141: Please change the comment from command specific to what it does and returns.

src/utils/versionNumber.ts
```diff
@@ -26,6 +26,23 @@ export class VersionNumber {
     public build: string | number
   ) {}
 
+  public static parseMajorMinor(versionString: string): { major: number; minor: number } {
```
shetzel, line 29: Add jsdoc.

shetzel (approved):
Just questions and nits.

src/package1/package1Version.ts
WillieRuemmele: no aliases for package1

peternhale (approved):

PR #63, opened 2022-09-29
refactor: change list function to be static

shetzel:
Changes the `list` function, which returns a list of matching PackageVersions given an array of 0Ho (or aliases) and filtering options, to be static and take additional args for Connection and SfProject.  This is necessary to adopt the PackageVersion refactoring effort in the plugin.

@W-11819142@

### DIFF
--- a/src/package/packageVersion.ts
+++ b/src/package/packageVersion.ts
@@ -231,6 +231,28 @@ export class PackageVersion {
     }
   }
 
+  public static async list(
+    connection: Connection,
+    project: SfProject,
+    options: PackageVersionListOptions
+  ): Promise<PackageVersionListResult[]> {
+    // resolve/verify packages
+    const packages = options.packages.map((pkg) => {
+      const id = getPackageIdFromAlias(pkg, project);
+
+      // validate ID
+      if (id.startsWith('0Ho')) {
+        validateId(BY_LABEL.PACKAGE_ID, id);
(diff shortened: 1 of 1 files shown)


peternhale (approved):

src/package/packageVersion.ts
```diff
@@ -231,6 +231,28 @@ export class PackageVersion {
...
+    connection: Connection,
+    project: SfProject,
+    options: PackageVersionListOptions
+  ): Promise<PackageVersionListResult[]> {
+    // resolve/verify packages
+    const packages = options.packages.map((pkg) => {
+      const id = getPackageIdFromAlias(pkg, project);
+
+      // validate ID
+      if (id.startsWith('0Ho')) {
```
WillieRuemmele, line 244: NIT: should we expand `validateId` to return the id if it's valid? it seems like we're starting to validate the id with `.startsWith` and then call `validateId` which seems to be doing the same job as the `if` above, might just make it more intuitive

WillieRuemmele (approved):

src/package/packageVersion.ts
shetzel: Yes I think ID validation logic can be improved.  That was work I started doing but put on hold because the API refactoring was changing the usecases.  It can be picked up again after we get the API more solid.

PR #78, opened 2022-10-07
fix: create subscriber package version class

peternhale:
@W-11798867@
Create a new SubscriberPackageVersion class that covers all installation related activities


src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+      'Package2ContainerOptions'
+    );
+    return this.Package2ContainerOptions;
+  }
+
+  public getPassword(): Optional<string> {
+    return this.password;
+  }
+
+  public async getSubscriberPackageId(): Promise<string> {
```
peternhale, line 240: All of the getters for SPV fields follow this pattern, except for getId. This pattern facilitates lazy field getting and value reuse.

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+  /**
+   * Get the SubscriberPackageVersion SObject data for this SubscriberPackageVersion.
+   *
+   * @param force - force a refresh of the subscriber package version data.
+   * @returns {PackagingSObjects.SubscriberPackageVersion} SObject data.
+   */
+  public async getData(
+    options: { force?: boolean; includeHighCostFields?: boolean } = { force: false, includeHighCostFields: false }
+  ): Promise<PackagingSObjects.SubscriberPackageVersion> {
+    if (!this.data || options.force || options.includeHighCostFields) {
```
peternhale, line 388: Additional check for high cost field has been requested that will cause a query to run

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+      const queryFields = this.getFieldsForQuery(options);
```
peternhale, line 389: narrows the query fields to only those that need to be queried. If the queryFields list is empty the current data are returned.

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+        const queryWithKey = `${queryNoKey} AND InstallationKey ='${escapedInstallationKey}'`;
+        this.data = await this.connection.singleRecordQuery<PackagingSObjects.SubscriberPackageVersion>(queryWithKey, {
+          tooling: true,
+        });
+      } catch (err) {
+        throw messages.createError('errorInvalidIdNoRecordFound', [this.options.id], undefined, err as Error);
+      }
+
+      // map the fields returned from the query to the class properties
+      this.mapFields(queryFields);
```
peternhale, line 405: House keeping to keep instance fields in sync with this.data

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+    const cspTrustedSites = await this.getCspTrustedSites();
+
+    const rssUrls = remoteSiteSettings?.settings ? remoteSiteSettings.settings?.map((rss) => rss.url) : [];
+    const cspUrls = cspTrustedSites?.settings ? cspTrustedSites?.settings.map((csp) => csp.endpointUrl) : [];
+
+    const sites = [...rssUrls, ...cspUrls];
+    return sites.length > 0 ? sites : undefined;
+  }
+
+  private async getFieldValue<T>(field: string): Promise<T> {
```
peternhale, line 483: This function takes a field name and then uses the field name to lookup the field in the instance. If not, the a request to get data for the missing field.

Downside to this is if the field has a falsy value, then a get on a falsy field value will cause another query. Suggestions welcome.

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+    if (!Reflect.has(this, field)) {
+      await this.getData({ includeHighCostFields: highCostQueryFields.includes(field) });
+    }
+    return Reflect.get(this, field);
+  }
+
+  private getFieldsForQuery(options: { force?: boolean; includeHighCostFields?: boolean }): string[] {
```
peternhale, line 490: Using the full list of fields provide of fields that need to be queried.
It resolves if high cost query fields should be included
It retains fields that not been read force === false
It retains fields if force === true

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+    const queryFields = SubscriberPackageVersionFields.filter(
+      (field) => !highCostQueryFields.includes(field) || options.includeHighCostFields
+    ).filter((field) => (!this.fieldsRead.has(field) && !options.force) || options.force);
+    return queryFields;
+  }
+
+  private mapFields(fields: string[]): void {
```
peternhale, line 497: For each field returned from a query, set the value in this.data and add to visited set.

src/package/package.ts
```diff
@@ -37,135 +41,218 @@ const packagePrefixes = {
...
-    super(options);
+    let packageId = this.options.packageAliasOrId;
+    if (!packageId.startsWith(packagePrefixes.PackageId)) {
+      packageId = getPackageIdFromAlias(this.options.packageAliasOrId, this.options.project);
+      if (packageId === this.options.packageAliasOrId) {
+        throw messages.createError('packageAliasNotFound', [this.options.packageAliasOrId]);
+      }
+    }
+
+    if (packageId.startsWith(packagePrefixes.PackageId)) {
```
WillieRuemmele, line 81: combine into `if {} else if {} else{}`?

src/package/package.ts
```diff
@@ -37,135 +41,218 @@ const packagePrefixes = {
...
-  }
+  /**
+   * Returns the list of package versions for the package.
+   * See {@link PackageVersionListOptions} for list options
+   *
+   * @param options
+   * @returns {Promise<PackageVersionListResult[]>}
+   */
+  public async getPackageVersions(options?: PackageVersionListOptions): Promise<PackageVersionListResult[]> {
+    // This should be calling PackageVersion.list() here, but that method is not implemented yet.
```
WillieRuemmele, line 191: is this comment still true?

src/package/packageVersionList.ts
```diff
@@ -35,6 +35,7 @@ const logger = Logger.childFromRoot('packageVersionList');
 export async function listPackageVersions(
   options: ListPackageVersionOptions
 ): Promise<QueryResult<PackageVersionListResult>> {
+  // TODO: replaces with connection.autoFetchQuery w/ tooling enabled
```
WillieRuemmele, line 38: still need to do this?

src/package/package.ts
peternhale: Comment is no longer accurate - removed.

src/package/packageVersionList.ts
peternhale: Still waiting on review https://github.com/forcedotcom/sfdx-core/pull/665

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+    }
+    return installRequest;
+  }
+
+  /**
+   * Get the package version ID for this PackageVersion.
+   *
+   * @returns The PackageVersionId (05i).
+   */
+  public async getId(): Promise<string> {
```
WillieRuemmele, line 219: why is this async?

peternhale: Does not need to be, but I will retain the promise return type for consistency across getters.

WillieRuemmele: ok, it's breaking an eslint rule, but I'll leave it be

src/package/packageVersion.ts
```diff
@@ -277,6 +349,26 @@ export class PackageVersion {
...
   }
 
+  public async getPackageId(): Promise<string> {
+    if (!this.data.Package2Id) {
+      await this.getPackageVersionData();
+    }
+    return this.data.Package2Id;
+  }
+
+  public async getPackageType(): Promise<PackageType> {
```
shetzel, line 359: `getPackageId` and `getPackageType` should have jsDoc.  `getPackageId` jsDoc should mention the returned ID prefix (0Ho).

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+    }
+    const installRequest = await getStatus(connection, installRequestId);
+    if (!installRequest) {
+      throw messages.createError('packageVersionInstallRequestNotFound', [installRequestId]);
+    }
+    return installRequest;
+  }
+
+  /**
+   * Get the package version ID for this PackageVersion.
```
shetzel, line 215:
```suggestion
   * Get the package version ID for this SubscriberPackageVersion.
```

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+};
+
+const allZeroesInstallOptions: PackageInstallOptions = {
+  pollingFrequency: Duration.minutes(0),
+  pollingTimeout: Duration.minutes(0),
+  publishFrequency: Duration.minutes(0),
+  publishTimeout: Duration.minutes(0),
+};
+
+export class SubscriberPackageVersion implements PackagingSObjects.SubscriberPackageVersion {
```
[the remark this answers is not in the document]
shetzel: I don't like all the public getters either.  Can't we just have a private object as an instance (e.g., `data`) that has all the fields and then have a public `getField(field: string)`?

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+  ): Promise<PackagingSObjects.PackageInstallRequest> {
+    try {
+      // before starting the install, check to see if the package version is available for install
+      await waitForPublish(
+        this.connection,
+        await this.getId(),
+        options.publishFrequency,
+        options.publishTimeout,
+        pkgInstallCreateRequest.Password
+      );
```
shetzel, line 434: Shouldn't `waitForPublish` only be run if an option is set?  Otherwise, the command parameter in the plugin is meaningless.

src/package/subscriberPackageVersion.ts
shetzel: Shouldn't `getFieldValue` check that `this.data?[field] != null` and return it if true, otherwise fetch it?

src/package/subscriberPackageVersion.ts
```diff
@@ -0,0 +1,503 @@
...
+  public RemoteSiteSettings: PackagingSObjects.SubscriberPackageRemoteSiteSettings;
+  public CspTrustedSites: PackagingSObjects.SubscriberPackageCspTrustedSites;
+  public Profiles: PackagingSObjects.SubscriberPackageProfiles;
+  public Dependencies: PackagingSObjects.SubscriberPackageDependencies;
+  public InstallValidationStatus: PackagingSObjects.InstallValidationStatus;
+  // end of fields from the SubscriberPackageVersion object
+  private readonly password: Optional<string>;
+  private readonly connection: Connection;
+  private data: PackagingSObjects.SubscriberPackageVersion;
+  private fieldsRead = new Set<string>();
```
shetzel, line 123: What is `fieldsRead`?

shetzel (changes requested):

src/package/subscriberPackageVersion.ts
peternhale: Tracks if a field has been queried.

src/package/packageVersionList.ts
peternhale: Replaced

src/package/subscriberPackageVersion.ts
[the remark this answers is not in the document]
peternhale: Reworked to a minimum number of field getters, made getField public and folks can use getData to access all of the data and use dot field refs to get values.

src/package/subscriberPackageVersion.ts
peternhale: I removed async and return a Promise.resolve...

src/package/subscriberPackageVersion.ts
peternhale: @shetzel the check to skip the wait is in the waitForPublish function, same with installStatus

src/package/subscriberPackageVersion.ts
peternhale: It will :)

WillieRuemmele (approved):

shetzel (approved):

PR #85, opened 2022-10-10
@W-11793337@: feat: added an option definitionFile to the convert command

smotwani11:


smotwani11:
@W-11793337@

Adds an option --definitionfile to the convert command

peternhale (changes requested):

src/utils/packageUtils.ts
```diff
@@ -74,6 +74,12 @@ export function validateId(idObj: Many<IdRegistryValue>, value: string): void {
     ]);
   }
 }
+
+export async function getApiVersion(project: SfProject): Promise<string> {
```
peternhale, line 78: @smotwani11 suggest that the function be renamed to `getSourceApiVersion`

peternhale (changes requested):

src/package/packageConvert.ts
```diff
@@ -136,26 +145,66 @@ export async function convertPackage(
...
   const packageVersBlobZipFile = path.join(packageVersTmpRoot, consts.PACKAGE_VERSION_INFO_FILE_ZIP);
 
   const packageDescriptorJson = {
     id: packageId,
   };
 
+  const settingsGenerator = new SettingsGenerator({ asDirectory: true });
+  const definitionFile = context.definitionfile;
+
+  if (definitionFile) {
```
peternhale, line 167: The package convert command can now fail due to additional causes, one that is tested for `signupDuplicateSettingsSpecified` and the other some sort of file i/o error or a json parse error. Both will be rather cryptic. Maybe add try/catch for this block and provide a wrapped error with a message that is under our control. the `messages.createError` function can take another error as a parameter, retaining the original cause.

peternhale (changes requested):
@smotwani11 I am afraid I am asking for another change around the processing of the definition file.

src/package/packageConvert.ts
```diff
@@ -136,29 +145,72 @@ export async function convertPackage(
...
 
+  await settingsGenerator.createDeploy();
+  await settingsGenerator.createDeployPackageContents(apiVersion);
+  await srcDevUtil.zipDir(
+    `${settingsGenerator.getDestinationPath()}${path.sep}${settingsGenerator.getShapeDirName()}`,
+    settingsZipFile
+  );
+
+  const shapeDirectory = `${settingsGenerator.getDestinationPath()}${path.sep}${settingsGenerator.getShapeDirName()}`;
+  const currentPackageXml = await fs.promises.readFile(path.join(shapeDirectory, 'package.xml'), 'utf8');
```
peternhale, line 204: @smotwani11 is package.xml file guaranteed to be at the location produced by the path.join? If not, the read should be in a try/catch.

test/package/packageConvert.test.ts
```diff
@@ -31,9 +37,27 @@ describe('packageConvert', () => {
   describe('createPackageVersionCreateRequest', () => {
     it('should return a valid request with installationkey and buildinstance', async () => {
       $$.inProject(true);
+      const definitionFile = {
```
peternhale, line 40: @smotwani11 given this addition test setup, is there a measurable effect that can be inspected for correctness?

src/package/packageConvert.ts
```diff
@@ -136,29 +145,72 @@ export async function convertPackage(
...
     id: packageId,
   };
 
+  const settingsGenerator = new SettingsGenerator({ asDirectory: true });
+  const definitionFile = context.definitionfile;
+
+  try {
+    if (definitionFile) {
+      const definitionFilePayload = await fs.promises.readFile(definitionFile, 'utf8');
+      const definitionFileJson = JSON.parse(definitionFilePayload) as ScratchOrgInfo;
```
peternhale, line 170: @smotwani11 how do you feel about placing a try/catch around the read and parse that throws the `errorReadingDefintionFile` error and then removing the block try/catch?

src/utils/packageUtils.ts
```diff
@@ -74,6 +74,12 @@ export function validateId(idObj: Many<IdRegistryValue>, value: string): void {
     ]);
   }
 }
+
+export async function getSourceApiVersion(project: SfProject): Promise<string> {
+  const apiVersion = project.getSfProjectJson().get('sourceApiVersion') as string;
+  return apiVersion;
+}
```
peternhale, line 81: This function does not need to be async
``` suggestion
export function getSourceApiVersion(project: SfProject): string {
  return project.getSfProjectJson().get('sourceApiVersion') as string;
}
```

src/package/packageConvert.ts
smotwani11: yes package.xml will be generated by the settings generator. The tests which don't provide the definition file option cover that scenario. The package.xml will be present in that location.

src/package/packageConvert.ts
smotwani11: I will resolve this.

test/package/packageConvert.test.ts
smotwani11: All the setup changes go into a zip file string representation:
Pointing you to the comment right above the expect in the same test:

// the most we can assert about VersionInfo because it is a zip file string representation, which changes with time
      expect(typeof request.VersionInfo).to.equal('string');

peternhale (approved):

peternhale (approved):

PR #101, opened 2022-10-20
fix: guard against empty component set

peternhale:
@W-11903179@

### DIFF
--- a/src/package/packageProfileApi.ts
+++ b/src/package/packageProfileApi.ts
@@ -229,7 +229,7 @@ export class PackageProfileApi extends AsyncCreatable<ProfileApiOptions> {
     const profilePaths = this.findAllProfiles(excludedDirectories);
 
     // Filter all profiles
-    typesArr = typesArr.filter((kvp) => kvp.name[0] !== 'Profile');
+    typesArr = (typesArr || []).filter((kvp) => kvp.name[0] !== 'Profile');
 
     if (profilePaths) {
       const members: string[] = [];
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -110,6 +110,7 @@ export class PackageVersionCreate {
       convertResult.packagePath = outputDirectory;
       return convertResult;
     }
+    return convertResult;
   }
 
   private async validateDependencyValues(dependency: PackageDescriptorJson): Promise<void> {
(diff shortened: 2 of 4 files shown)


src/package/packageProfileApi.ts
```diff
@@ -229,7 +229,7 @@ export class PackageProfileApi extends AsyncCreatable<ProfileApiOptions> {
     const profilePaths = this.findAllProfiles(excludedDirectories);
 
     // Filter all profiles
-    typesArr = typesArr.filter((kvp) => kvp.name[0] !== 'Profile');
+    typesArr = (typesArr || []).filter((kvp) => kvp.name[0] !== 'Profile');
```
peternhale, line 232: Make sure value passed is not undefined

src/package/packageVersionCreate.ts
```diff
@@ -110,6 +110,7 @@ export class PackageVersionCreate {
       convertResult.packagePath = outputDirectory;
       return convertResult;
     }
+    return convertResult;
```
peternhale, line 113: Fix return to always return the converted component set

src/package/packageVersionCreate.ts
```diff
@@ -110,6 +110,7 @@ export class PackageVersionCreate {
       convertResult.packagePath = outputDirectory;
       return convertResult;
```
peternhale, line 111: I will remove this return

src/package/packageVersionCreate.ts
```diff
@@ -419,6 +421,12 @@ export class PackageVersionCreate {
     return this.createRequestObject(preserveFiles, packageVersTmpRoot, packageVersBlobZipFile);
   }
 
+  private verifyHasSource(componentSet: ConvertResult): void {
```
peternhale, line 424: Verify that the component set has source present
Wrote a function so it was simple to mock

src/package/packageVersionCreate.ts
```diff
@@ -437,6 +445,12 @@ export class PackageVersionCreate {
     const currentPackageXml = await fs.promises.readFile(path.join(packageVersMetadataFolder, 'package.xml'), 'utf8');
     // convert to json
     const packageJson = await xml2js.parseStringPromise(currentPackageXml);
+    if (!packageJson?.Package) {
```
peternhale, line 448: Unlikely to happen now, but leaving here just in case

shetzel (approved):

WillieRuemmele (approved):

PR #109, opened 2022-10-24
@W-11824088@ Enable support for language to be supplied to version creation and ve…

sethmuss:
…rsion display

@W-11824088@

### DIFF
--- a/src/interfaces/packagingInterfacesAndType.ts
+++ b/src/interfaces/packagingInterfacesAndType.ts
@@ -134,6 +134,7 @@ export type PackageVersionListResult = {
   ReleaseVersion?: string;
   BuildDurationInSeconds?: number;
   HasMetadataRemoved?: boolean;
+  Language?: string;
 };
 
 export type PackageInstallCreateRequest = Partial<
@@ -185,6 +186,7 @@ export type PackageDescriptorJson = Partial<NamedPackageDir> &
     subscriberPackageVersionId: string;
     packageId: string;
     versionName: string;
+    language: string;
   }>;
 
 export type PackageVersionCreateRequest = {
@@ -195,6 +197,7 @@ export type PackageVersionCreateRequest = {
   InstallKey: string;
   Instance: string;
   SourceOrg: string;
+  Language: string;
   CalculateCodeCoverage: boolean;
   SkipValidation: boolean;
 };
@@ -207,6 +210,7 @@ export type PackageVersionListOptions = {
   verbose: boolean;
   concise: boolean;
   isReleased: boolean;
+  connection: Connection;
 };
 
 export type PackageVersionUpdateOptions = {
@@ -283,6 +287,7 @@ export type PackageVersionCreateOptions = {
   definitionfile: string;
   installationkey: string;
   installationkeybypass: boolean;
+  language: string;
   packageId: string;
   path: string;
   postinstallscript: string;
--- a/src/package/packageVersionList.ts
+++ b/src/package/packageVersionList.ts
@@ -28,6 +28,13 @@ const VERBOSE_SELECT =
   'ConvertedFromVersionId, Package2.IsOrgDependent, ReleaseVersion, BuildDurationInSeconds, HasMetadataRemoved, CreatedById ' +
   'FROM Package2Version';
 
+const VERBOSE_SELECT_57 =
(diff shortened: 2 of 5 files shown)


SagarBijjam (approved):

src/package/packageVersionList.ts
```diff
@@ -28,6 +28,13 @@ const VERBOSE_SELECT =
   'ConvertedFromVersionId, Package2.IsOrgDependent, ReleaseVersion, BuildDurationInSeconds, HasMetadataRemoved, CreatedById ' +
   'FROM Package2Version';
 
+const VERBOSE_SELECT_57 =
```
peternhale, line 31: @sethmuss is there any other way the new field can be isolated other than creating a query specific for that field?

peternhale (changes requested):

src/interfaces/packagingInterfacesAndType.ts
```diff
@@ -283,6 +287,7 @@ export type PackageVersionCreateOptions = {
   definitionfile: string;
   installationkey: string;
   installationkeybypass: boolean;
+  language: string;
```
peternhale, line 290: The location field needs to be added to the appropriate sobject found in packagingSObjects.ts

peternhale (changes requested):

src/package/packageVersionList.ts
shetzel: We can use arrays of fields to build the queries to keep it DRY.  E.g., 
```
const defaultFields = ['Id', 'Package2Id', ...];
const verboseFields = ['CodeCoverage', ...];
const verbose57Fields = ['Language'];
let queryFields = defaultFields;
if (options.verbose) {
  queryFields = [...queryFields, ...verboseFields];
  if (connection.version >= 57) {
    queryFields = [...queryFields, ...verbose57Fields];
  }
}
const query = `SELECT ${queryFields.toString()} FROM Package2Version`;
```

The only other way I can think to handle this API version specific code is to assume the field exists, wrap the query in a try/catch and if it fails display a warning message that they can't use the Language feature unless they update to v57.0+, then send the more basic verbose query.

sethmuss: But we really don't need all that right? As soon as App server 242 rolls out we'll move the current  VERBOSE_SELECT_57 to VERBOSE_SELECT. I think logical DRY: great. But it's just an extra static field.

I'll admit part of the reason I've written it the way I have is I'm not confident with javascript :)  Like: 
"queryFields = [...queryFields, ...verboseFields];" no idea what those "..."'s are nor the rules around literal arrays(?), and what the gotchas might be. I think if the packaging team is going to own this stuff either: we need to invest in becoming javascript experts, or, keep things as simple as possible.

src/interfaces/packagingInterfacesAndType.ts
sethmuss: Thank you @peternhale . Not sure how I missed that in my local repo.

sethmuss: Done

src/package/packageVersionList.ts
peternhale: @sethmuss Sorry for the delay getting back to you.

 IIRC, when the field was added it carried the attribute of which version of app server the field was introduced. So FLS is going to reject requests using this field if a user uses an API call with a version < 242. 

Having to account for API version backwards compatibility in this way is not optimal.

I have to admit that version handling for FLS at the sobject level in CLI libraries and even jsforce are almost nonexistent, probably for the same reason that started this discussion.

These are the options:
* Like Steve's example, isolate the knowledge of the new field being version sensitive and use it only when api version is suitable
* try/catch - examine the exception for the appropriate message, adjust the request and retry
* Fetch the sobject describe. The results will only include fields that one has access to, so the request can then be tailored to those fields (jsforce already supports this, so getting it to work is not difficult).
* Use the pattern in the PR
* Wait to merge this PR into main until 242 R2. We have the capability to produce prerelease version of the plugin so your team can do the needed testing before R2.

**Ultimately this will be the packaging teams decision, because it will be yours' to maintain.**

Yes, the team will need to get additional training on javascript and typescript.

We are happy to pair with you and other team members if you need help.

PR #120, opened 2022-11-04
fix: remove objects that cause circular errors in logging

peternhale:
@W-12010479@

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -728,7 +728,10 @@ export class PackageVersionCreate {
               versionNumber.indexOf(pkgUtils.VERSION_NUMBER_SEP + BuildNumberToken.NEXT_BUILD_NUMBER_TOKEN)
             )
           : versionNumber;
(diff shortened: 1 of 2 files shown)


src/package/packageVersionCreate.ts
```diff
@@ -728,7 +728,10 @@ export class PackageVersionCreate {
               versionNumber.indexOf(pkgUtils.VERSION_NUMBER_SEP + BuildNumberToken.NEXT_BUILD_NUMBER_TOKEN)
             )
           : versionNumber;
-      this.logger.warn(options, messages.getMessage('defaultVersionName', [packageDescriptorJson.versionName]));
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { connection: c, project: p, profileApi: a, ...warnOptions } = options;
```
peternhale, line 733: The circular reference when trying to stringify is coming from the connection instance. Removed it and two other classes to be sure the serialization is clean.

PR #154, opened 2022-12-10
Language lib @W-11824088@

sethmuss:
@W-11824088@ Add language to package version commands: create and list.

Prev PR: https://github.com/forcedotcom/packaging/pull/109/files

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -303,6 +303,7 @@ export class PackageVersionCreate {
       SourceOrg: this.options.sourceorg,
       CalculateCodeCoverage: this.options.codecoverage || false,
       SkipValidation: this.options.skipvalidation || false,
+      Language: this.options.language,
     };
 
     if (preserveFiles) {
@@ -396,8 +397,9 @@ export class PackageVersionCreate {
     // (see _retrieveSubscriberPackageVersionId for details)
     const dependencies = packageDescriptorJson.dependencies;
 
-    // branch can be set via options or descriptor; option takes precedence
+    // branch and language can be set via options or descriptor; option takes precedence
     this.options.branch = this.options.branch ?? packageDescriptorJson.branch;
+    this.options.language = this.options.language ?? packageDescriptorJson.language;
 
     const resultValues = await Promise.all(
       !dependencies ? [] : dependencies.map((dependency) => this.retrieveSubscriberPackageVersionId(dependency))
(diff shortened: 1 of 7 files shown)


src/package/packageVersionCreate.ts
```diff
@@ -303,6 +303,7 @@ export class PackageVersionCreate {
       SourceOrg: this.options.sourceorg,
       CalculateCodeCoverage: this.options.codecoverage || false,
       SkipValidation: this.options.skipvalidation || false,
+      Language: this.options.language,
```
peternhale, line 306: @sethmuss does including language in the request need to be guarded with api version?

src/package/packageVersionList.ts
```diff
@@ -14,19 +14,40 @@ import { PackageVersionListResult, PackageVersionListOptions, ListPackageVersion
...
-  'CreatedDate, LastModifiedDate, IsPasswordProtected, AncestorId, ValidationSkipped, CreatedById ' +
-  'FROM Package2Version';
-
-const VERBOSE_SELECT =
-  'SELECT Id, Package2Id, SubscriberPackageVersionId, Name, Package2.Name, Package2.NamespacePrefix, ' +
-  'Description, Tag, Branch, MajorVersion, MinorVersion, PatchVersion, BuildNumber, IsReleased, ' +
-  'CreatedDate, LastModifiedDate, IsPasswordProtected, CodeCoverage, HasPassedCodeCoverageCheck, AncestorId, ValidationSkipped, ' +
-  'ConvertedFromVersionId, Package2.IsOrgDependent, ReleaseVersion, BuildDurationInSeconds, HasMetadataRemoved, CreatedById ' +
-  'FROM Package2Version';
+const defaultFields = [
```
peternhale, line 17: Thank you for doing this.

peternhale (changes requested):

src/package/packageVersionCreate.ts
sethmuss: @peternhale Good question. I went with no because the app server gives a decent error: "No such column 'Language' on sobject of type Package2VersionCreateRequest" 

If you think it's better to have a more specific error I can add. Let me know what you think!

sethmuss: (that error is only passed back if the user tries to send in a language via sfdx-project.json or cmd line param)

peternhale: @sethmuss On the command side, will the flag to specify language have a default value?

sethmuss: @peternhale Yes. The app server defaults the language value to the auth'd dev hub user's language. For backward compatibility and a couple other reasons.

peternhale (approved):

azenoni (approved):

PR #186, opened 2023-01-09
fix: search for pkg dir by pkg id

peternhale:
@W-12349329@

- modify validation to first find a package in package dirs if 0Ho is provided
- capture the id as the packageName
- If more than one pkg dir is found that references the same id, defer to alias lookups


WillieRuemmele (approved):

shetzel (approved):

WillieRuemmele:
### QA Notes 
---

✅ : verified that setting a `PackageDirectory.package` to an 'OH0' value worked
✅ : `package` as an alias with a single entry
✅ : `package` as an alias with multiple entries

PR #192, opened 2023-01-17
fix: do not add language from org descriptor to pvc request object

shetzel:
If a language was defined in an org file descriptor used for a package version create, it would set that as an option on the package version create request.  This fix removes that.

Also, the language option is removed unless an org connection of API version 57.0 or higher is used.

@W-12377800@

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -309,8 +309,7 @@ export class PackageVersionCreate {
 
     // Ensure we only include the Language property for a connection api version
     // of v57.0 or higher.
-    const apiVersion = this.connection.getApiVersion();
-    if (apiVersion && apiVersion < '57.0') {
+    if (this.connection.getApiVersion() < '57.0') {
       if (requestObject.Language) {
         this.logger.warn(
           `The language option is only valid for API version 57.0 and higher. Ignoring ${requestObject.Language}`
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -307,6 +307,18 @@ export class PackageVersionCreate {
       Language: this.options.language,
     };
 
+    // Ensure we only include the Language property for a connection api version
+    // of v57.0 or higher.
+    const apiVersion = this.connection.getApiVersion();
+    if (apiVersion && apiVersion < '57.0') {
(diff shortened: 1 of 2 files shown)


src/package/packageVersionCreate.ts
```diff
@@ -307,6 +307,18 @@ export class PackageVersionCreate {
       Language: this.options.language,
     };
 
+    // Ensure we only include the Language property for a connection api version
+    // of v57.0 or higher.
+    const apiVersion = this.connection.getApiVersion();
+    if (apiVersion && apiVersion < '57.0') {
```
WillieRuemmele, line 313: `if (this.connection.getApiVersion() < '57.0')`

WillieRuemmele (approved):

sethmuss (approved):

WillieRuemmele (approved):

SKuhnt:
Hi @shetzel,

I have a Problem/Question regarding:
> Also, the language option is removed unless an org connection of API version 57.0 or higher is used.

Does that mean I can't use the language option on version create when my DevHub Org is still on version 56? (or rather it is not applied)
When I'm not using the language option for some reason my Package is generating German metadata. Thus getting an Error.
e.g. 
> Field: Cannot modify managed object: entity=CustomFieldDefinition, [...] Datatype Previous Value: Picklist (Multi-Select) New Value: Auswahlliste (Mehrfachauswahl)

I tried uninstalling cli and installing an Old version (7.177.1) but it is not working and I'm still getting the same Error.
Do you have any suggestions what i might do to handle this?

Also I'm wondering why is package version create trying to do it in german anyway?
My Mac is setup in english, my Salesforce User is set to english as language and also the Org language is english.
(Only the Locale is German)

any Help is appreciated.

shetzel:
@SKuhnt 

> Does that mean I can't use the language option on version create when my DevHub Org is still on version 56? (or rather it is not applied)
Correct.  The `Language` field was added to the `PackageVersionCreateRequest` in API version 57.0 (Spring '23) so it would error anyway if you tried setting it pre v57.0.  To be clear, this is only about the language field on the `PackageVersionCreateRequest`, not the language property you can set in your org config file.  They are separate things.  If you need to use the language property in your org config file then continue to do that.

PR #201, opened 2023-01-24
Ts/seed metadata Added support for seedMetadata to package version create and convert commands

tshopshireSalesforce:
I added support for seedMetadata to the convert and package version create commands. Seed metadata will be used for unpackageable metadata that needs to be installed on an org prior to deploying the packaged metadata.


tshopshireSalesforce:
Will recreate with my email address correctly associated.

PR #227, opened 2023-02-17
feat: for expandability, PackageInstallRequest should be a string value

KSCCO:
@W-12394717@ PackageInstallRequest SkipHandlers should be a string value in order to accommodate the potential for a list of handlers to skip in the future.

### DIFF
--- a/src/interfaces/packagingSObjects.ts
+++ b/src/interfaces/packagingSObjects.ts
@@ -267,7 +267,7 @@ export namespace PackagingSObjects {
     EnableRss: boolean;
     UpgradeType: Nullable<'delete-only' | 'deprecate-only' | 'mixed-mode'>;
     ApexCompileType: Nullable<'all' | 'package'>;
-    SkipHandlers: boolean;
+    SkipHandlers: string;
     Status: 'ERROR' | 'IN_PROGRESS' | 'SUCCESS' | 'UNKNOWN';
     Errors: Nullable<SubscriberPackageInstallErrors>;
   };


src/interfaces/packagingSObjects.ts
```diff
@@ -267,7 +267,7 @@ export namespace PackagingSObjects {
     EnableRss: boolean;
     UpgradeType: Nullable<'delete-only' | 'deprecate-only' | 'mixed-mode'>;
     ApexCompileType: Nullable<'all' | 'package'>;
-    SkipHandlers: boolean;
+    SkipHandlers: string;
```
peternhale, line 270: @KSCCO The contents of packagingSObjects.ts is meant to reflect the actual shape of the server side objects. You had mentioned that this property is actually a string array on the server side. If this is correct, then this should be a string[].

peternhale (changes requested):

src/interfaces/packagingSObjects.ts
KSCCO: @peternhale, the corresponding server side object is just a string but does get parsed as a comma separated string. 
https://sourcegraph.soma.salesforce.com/perforce.soma.salesforce.com/app/main/core@HEAD/-/blob/udd-xml/java/resources/udd/PackageInstallRequest.entity.xml?L42

peternhale: Thanks for clarifying @KSCCO.

peternhale (approved):

PR #234, opened 2023-02-22
fix: adds support for seed metadata to both the convert and package version create commands

tshopshireSalesforce:
What does this PR do?

Fixes package version create so profiles are excluded from the unpackaged metadata directory if code coverage is not specified.

@W-12572895@

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -486,7 +486,7 @@ export class PackageVersionCreate {
       {
         Package: typesArr,
       },
-      [unpackagedMetadataPath]
+      [hasUnpackagedMetadata ? unpackagedMetadataPath : null]
     );
 
     if (excludedProfiles.length > 0) {


src/package/packageVersionCreate.ts
```diff
@@ -486,7 +486,7 @@ export class PackageVersionCreate {
       {
         Package: typesArr,
       },
-      [unpackagedMetadataPath]
+      [hasUnpackagedMetadata ? unpackagedMetadataPath : null]
```
peternhale, line 489: @tshopshireSalesforce while this assignment is safer, the issue is the guard on setting var `hasUnpackagedMetadata` on line 467. Having unpackagedMetadata is independent of running tests with code coverage.

peternhale (changes requested):

src/package/packageVersionCreate.ts
ganeshnt: my obseravation agrees with @tshopshireSalesforce saying.  unpackagedMetadata worked only when code coverage option was provided.  now, am not sure what is the expected behavior though.  That someone from packaging can confirm

KSCCO: Here's my summary (correct me if anything here is wrong):
Tyler's original change caused the regression because it created a discrepancy between the profiles in the manifest 
 and the actual profiles, when "hasUnpackagedMetadata" is identified as false due to codecoverage not being specified. In that case, the unpackageMetadataPath is **not** excluded from the manifest creation, but the path is excluded from the profile generation.
 
This fix should resolve that. 

It's possible I'm missing something, but I would think that this logic should be agnostic to whether or not the codecoverage flag is used. I understand that the unpackaged metadata dir is designed for adding metadata needed to run Apex tests and that Apex tests don't run unless the codecoverage flag is specified, but why would we want to allow profiles from a directory identified as being used for unpackaged metadata to be included in the package profiles? It's likely relatively innocent since you cannot release a package that hasn't run codecoverage.

My suggestion would be to procced with this change and I'll raise this with unlocked packaging team to handle with a separate lower priority bug.

KSCCO (approved):

peternhale (approved):

PR #239, opened 2023-03-01
fix: when creating version, set language of build org from scratch or…

t7job:
…g definition file's language

@W-12513253@

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -254,6 +254,7 @@ export class PackageVersionCreate {
       SourceOrg: this.options.sourceorg,
       CalculateCodeCoverage: this.options.codecoverage || false,
       SkipValidation: this.options.skipvalidation || false,
+      // note: the createRequest's Language corresponds to the AllPackageVersion's language
       Language: this.options.language,
     };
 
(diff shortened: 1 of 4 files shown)


src/package/packageVersionCreate.ts
```diff
@@ -344,14 +353,17 @@ export class PackageVersionCreate {
         throw messages.createError('signupDuplicateSettingsSpecified');
       }
 
-      ['country', 'edition', 'features', 'orgPreferences', 'snapshot', 'release', 'sourceOrg'].forEach((prop) => {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-        const propValue = definitionFileJson[prop];
-        if (propValue) {
+      // these scratch org definition file values will be applied to the build org
+      ['country', 'language', 'edition', 'features', 'orgPreferences', 'snapshot', 'release', 'sourceOrg'].forEach(
```
shetzel, line 357: @sethmuss - this adds back the `language` from the scratch org def file like it was before when we had that bug (before 242 R2).  I thought you said that the language set for the scratch org should not be applied here?

sethmuss: @shetzel Yes. I said that because I didn't fully understand the use-case for it to be present in the scratch/build org for a package version create. A customer ran into an issue exposing that having the language in the scratch/build org is something we still need to support. This looks right to me. We have language on the Package2VersionCreateRequest on line 258, and language into the descriptor for the scratch/build org support. Sorry for the confusion Steve, that's my bad.

sethmuss (approved):

src/package/packageVersionCreate.ts
shetzel: Ok thanks for the verification @sethmuss!

shetzel (approved):

PR #277, opened 2023-04-22
fix: properly chunk query for getPackageVersionStrings

shetzel:
Fixes the getPackageVersionStrings() function to properly chunk a large query into multiple, smaller queries when the where clause is too large.

@W-13083719@

### DIFF
--- a/test/utils/packageUtils.test.ts
+++ b/test/utils/packageUtils.test.ts
@@ -9,13 +9,14 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import { assert, expect } from 'chai';
-import { instantiateContext, restoreContext, stubContext } from '@salesforce/core/lib/testSetup';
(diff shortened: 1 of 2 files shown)


mshanemc (approved):

test/utils/packageUtils.test.ts
```diff
@@ -67,7 +69,22 @@ describe('packageUtils', () => {
...
+        records: [{ MajorVersion: 1 }],
+        done: true,
+        totalSize: 1,
+      });
+
+      // generate a large array of fake subscriber package version IDs
+      const spvs: string[] = [];
+      while (spvs.length < 201) {
+        spvs.push($$.uniqid());
+      }
```
mshanemc, line 84:
```suggestion
      // generate a large array of fake subscriber package version IDs
      Array.from({ length: 201 }, () => $$.uniqid())
```

mshanemc (approved):

PR #283, opened 2023-05-03
feat: add AppAnalyticsEnabled option to PackageUpdateOptions

stephen-carter-at-sf:
This is the prerequisite for https://github.com/salesforcecli/plugin-packaging/pull/325

### DIFF
--- a/src/interfaces/packagingInterfacesAndType.ts
+++ b/src/interfaces/packagingInterfacesAndType.ts
@@ -48,6 +48,7 @@ export type PackageUpdateOptions = {
   Name?: string;
   Description?: string;
   PackageErrorUsername?: string;
+  AppAnalyticsEnabled?: boolean;
 };
 
 export type PackageIdType =
--- a/src/package/package.ts
+++ b/src/package/package.ts
@@ -51,6 +51,7 @@ export const Package2Fields = [
   'IsOrgDependent',
   'ConvertedFromPackageId',
   'PackageErrorUsername',
+  'AppAnalyticsEnabled',
 ];
 
 /**
@@ -116,7 +117,7 @@ export class Package {
   public static async list(connection: Connection): Promise<PackagingSObjects.Package2[]> {
     return (
       await connection.tooling.query<PackagingSObjects.Package2>(
-        `select ${Package2Fields.toString()} from Package2 ORDER BY NamespacePrefix, Name`,
+        `select ${this.getPackage2Fields(connection).toString()} from Package2 ORDER BY NamespacePrefix, Name`,
         {
           autoFetch: true,
           maxFetch: 10_000,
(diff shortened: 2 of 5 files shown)


src/interfaces/packagingInterfacesAndType.ts
```diff
@@ -48,6 +48,7 @@ export type PackageUpdateOptions = {
   Name?: string;
   Description?: string;
   PackageErrorUsername?: string;
+  AppAnalyticsEnabled?: boolean;
```
peternhale, line 51: @stephen-carter-at-sf this property should also be added to the Package2 type [here](https://github.com/forcedotcom/packaging/blob/main/src/interfaces/packagingSObjects.ts#L11).

src/package/package.ts
```diff
@@ -51,7 +51,7 @@ export const Package2Fields = [
   'IsOrgDependent',
   'ConvertedFromPackageId',
   'PackageErrorUsername',
-];
+]; // 'AppAnalyticsEnabled' is only available starting in api version 59.0. See getPackage2Fields below.
```
peternhale, line 54: @stephen-carter-at-sf I would suggest adding the field to the list now and then exclude it in function `this.getPackage2Fields` if API version < 59.

src/package/package.ts
```diff
@@ -193,6 +193,14 @@ export class Package {
     return convertPackage(pkgId, connection, options, project);
   }
 
+  private static getPackage2Fields(connection: Connection): string[] {
+    const fields = Package2Fields;
+    if (connection.getApiVersion() >= '59.0') {
+      fields.push('AppAnalyticsEnabled');
+    }
+    return fields;
```
peternhale, line 201:
``` suggestion
    const apiVersion = connection.getApiVersion();
    return Package2Fields.filter((field) => apiVersion >= '59.0' ? true : field !== 'AppAnalyticsEnabled');
```

src/package/package.ts
```diff
@@ -258,6 +266,10 @@ export class Package {
         Object.entries(options).filter(([, value]) => value !== undefined)
       ) as PackageUpdateOptions;
 
+      if (opts.AppAnalyticsEnabled !== undefined && this.options.connection.getApiVersion() < '59.0') {
```
peternhale, line 269: @stephen-carter-at-sf have you considered just letting the update fail and deal with the error in the catch block, if you want provide a better message, that is.

peternhale (changes requested):

src/package/package.ts
stephen-carter-at-sf: The suggestion to let it fail and catch it goes against my instinctive defensive programming mentality. I've found its best to assert state before attempting any code that has mutation involved. That is, I wouldn't want to perform an update and then let it fail because it trusts the update function too much in my opinion. Also my guess is that the catch block would catch generic exception types that would require analysis to map it back to the correct error message. So I think I like this assertion before update better.

peternhale (approved):
@stephen-carter-at-sf Changes look good to me. Please include someone from the managed packaging team in this review. This is their library and I would like their approval.

sethmuss:
@stephen-carter-at-sf What's the WI for this work? Is there an associated app server WI?

stephen-carter-at-sf:
> @stephen-carter-at-sf What's the WI for this work? Is there an associated app server WI?

@sethmuss The work item is W-13112782 and the associated core side work item is W-13004068

sethmuss (approved):

shetzel (approved):

PR #295, opened 2023-05-19
feat: Package version retrieve operation (W-12557202)

andrewslater:
- Add method to download metadata for a given package version (1GP or 2GP)
- Use multipart/form-data when creating new package versions (allows larger package sizes to work)


package.json
```diff
@@ -51,7 +51,7 @@
     "graphology-traversal": "^0.3.1",
     "graphology-types": "^0.24.7",
     "js2xmlparser": "^4.0.2",
-    "jsforce": "2.0.0-beta.22",
+    "jsforce": "2.0.0-beta.23",
```
andrewslater, line 54: Necessary for the new `encoding` option in jsforce so we can consume the binary response stream for the metadata zip.

src/utils/packageUtils.ts
```diff
@@ -47,6 +47,9 @@ export type IdRegistryValue = { prefix: string; label: string };
 export type IdRegistry = {
   [key: string]: IdRegistryValue;
 };
+export type UnzipResults = {
```
andrewslater, line 50: This smelled funny but I read that having an async function return a boolean promise was bad (don't remember the reason unfortunately – please set me straight if this is dumb) and at one point I was thinking about including a manifest of unzipped files in the return value – which I clearly didn't end up doing so the value of this object is more iffy.

src/package/package.ts
```diff
@@ -194,6 +208,64 @@ export class Package {
...
+  /**
+   * Download the metadata files for a previously published package version, convert them to source format, and put them into a new project folder within the sfdx project.
+   *
+   * @param project
+   * @param options {@link PackageVersionMetadataDownloadOptions}
+   * @param connection
+   * @returns
+   */
+  public static async downloadPackageVersionMetadata(
+    project: SfProject,
```
shetzel, line 220: Does this really need a project?  Doesn't it just need an org connection and a 04t ID (and destination dir)?

src/package/package.ts
```diff
@@ -194,6 +208,64 @@ export class Package {
...
+    if (!versionInfo.MetadataZip) {
+      throw messages.createError('unableToAccessMetadataZip');
+    }
+
+    const responseBase64 = await connection.tooling.request<string>(versionInfo.MetadataZip, {
+      encoding: 'base64',
+    });
+    const buffer = Buffer.from(responseBase64, 'base64');
+
+    // Unzip the metadata zip file into a tmp directory. It still needs to be converted to source format.
```
shetzel, line 249: We should be able to take the response, build a ComponentSet from it in memory, convert it and then write to disk like the SDR library does.  Writing to disk twice should not be necessary.

src/package/packageVersionCreate.ts
```diff
@@ -659,7 +659,15 @@ export class PackageVersionCreate {
...
-    const createResult = await this.connection.tooling.create('Package2VersionCreateRequest', request);
+    const options = {
+      multipartFileFields: {
+        VersionInfo: {
+          contentType: 'application/zip',
+          filename: 'package-version-info.zip',
+        },
+      },
+    };
+    const createResult = await this.connection.tooling.create('Package2VersionCreateRequest', request, options);
```
shetzel, line 670: This needs to happen with every package version create?  I.e., it's not optional?  Is there a performance impact?

src/utils/packageUtils.ts
shetzel: Yeah, I would either return a manifest of files or some jsZip thing.  I'd have to look at the SDR code that handles retrieves but that general pattern should be followed to keep it all in memory until the source format files are written.

shetzel (changes requested):

src/interfaces/packagingSObjects.ts
```diff
@@ -344,5 +344,6 @@ export namespace PackagingSObjects {
     PatchVersion: number;
     BuildNumber: number;
     IsDeprecated: boolean;
+    MetadataZip: string;
```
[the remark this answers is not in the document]
andrewslater: It is available in API version 59+ and requires a specific user permission to access, so I'll mark it as optional.

src/package/package.ts
```diff
@@ -194,6 +208,64 @@ export class Package {
...
+   * @param connection
+   * @returns
+   */
+  public static async downloadPackageVersionMetadata(
+    project: SfProject,
+    options: PackageVersionMetadataDownloadOptions,
+    connection: Connection
+  ): Promise<PackageVersionMetadataDownloadResult> {
+    // Validate the destination path is suitable to extract package version metadata (must be new or empty)
+    const destinationFolder = options.destinationFolder || 'force-app/';
```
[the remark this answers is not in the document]
andrewslater: Our core use case for this feature is:
1. User creates new sfdx project
2. User converts their 1GP -> 2GP (may happen many times, but makes not changes to the project sources on local disk)
3. User uses `package version retrieve` command to download metadata for a converted 2GP. 
4. User uses retrieved metadata to create non-converted 2GP versions going forward

The idea would be that in such a happy path scenario the user doesn't provide `destinationFolder` and the metadata is retrieved into the standard `force-app` directory. There's a problem though, a "pristine" force-app directory tree still has empty folders and a couple of `.eslint.json` files. I want to change the utility method from `isDirEmpty` to `isPackageDirEffectivelyEmpty` which accounts for empty directories and `.eslint.json` files.

src/package/package.ts
```diff
@@ -194,6 +208,64 @@ export class Package {
...
+    const destinationPath = path.join(project.getPath(), destinationFolder);
```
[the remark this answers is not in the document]
andrewslater: Possibly confusing as it would make a relative directory inside their project dir. I suppose we could raise an error if the path starts with `path.sep`.

src/package/package.ts
andrewslater: At the moment technically it does not but we expect we'll want to introduce some quality of life enhancements that update `sfdx-project.json` after downloading the metadata (to better support the happy path use case I described in an earlier comment).

src/package/packageVersionCreate.ts
andrewslater: Technically if the payload is <25mb it does not need to use a mulitpart request. I considered branching the behavior but there should be no meaningful performance impact for using multipart requests in all cases so I opted for keeping it simple.

src/utils/packageUtils.ts
andrewslater: Removed entirely since we're not processing the zip in memory.

messages/package.md
```diff
@@ -25,3 +25,18 @@ A package with id %s was not found.
...
 Enabling App Analytics is only possible with API version 59.0 or higher.
+
+# sourcesDownloadDirectoryNotEmpty
+
+Can’t retrieve package version metadata. The specified directory isn’t empty. Empty the directory, or create a new one and then retry this command.
+
+# unableToAccessMetadataZip
+
+Can’t retrieve package version metadata. Ensure that you’re using API version 59.0 or higher, and that you have the user permissions needed to access fields on the MetadataPackageVersion object.
+Then retry this command.
```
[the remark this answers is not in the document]
andrewslater: Worked with doc writers to revise the language.

src/interfaces/packagingInterfacesAndType.ts
```diff
@@ -223,6 +224,13 @@ export type PackageVersionCreateRequestOptions = {
   skipancestorcheck?: boolean;
 };
 
+export type PackageVersionMetadataDownloadOptions = {
+  allPackageVersionId: string;
+  destinationFolder: string;
```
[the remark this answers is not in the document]
andrewslater: Changed to `subscriberPackageVersionid`

src/package/package.ts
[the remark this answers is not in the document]
andrewslater: Now raising an error if an absolute path is given

sethmuss (approved):
Looks good to me.

sethmuss (approved):
Looks good to me.

src/utils/packageUtils.ts
```diff
@@ -516,3 +516,28 @@ export function copyDescriptorProperties(
...
+  const entries = fs.readdirSync(directory, { withFileTypes: true });
+  let isEffectivelyEmpty = true;
+  entries.forEach((entry) => {
+    if (entry.isDirectory()) {
+      isEffectivelyEmpty = isPackageDirectoryEffectivelyEmpty(join(directory, entry.name));
+    } else if (entry.name !== '.eslint.json') {
+      isEffectivelyEmpty = false;
+    }
+  });
+  return isEffectivelyEmpty;
```
shetzel, line 542: nit: using `entries.every()` would short circuit faster with a bit less code. Lines 534-542 could be:
```typescript
return entries.every((entry) =>
  entry.isDirectory() ? isPackageDirectoryEffectivelyEmpty(join(directory, entry.name)) : entry.name === '.eslint.json')
);
```

messages/package.md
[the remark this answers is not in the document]
shetzel: "Then retry this command." is still in the message.

shetzel (changes requested):

PR #310, opened 2023-05-31
fix: when resolving dependencies the branch attribute should be preferred

David-Polehonski:
When resolving package version dependencies the branch attribute of the dependency should be used to select the package version, not the branch name in the parameter unless the dependency doesn't specify a branch.

For example, my dependency below failed to resolve when building a package version because an equivalent build number (0.2.4.1) doesn't exist on branch `dev` when running the command:
`sf package version create -p myPackage -b dev -k redacted -w 30 --code-coverage`
`"dependencies": [
        {
          "package": "myPackageDependency",
          "versionNumber": "0.2.4.1",
          "branch": "build"
        },`

I have built and linked packages locally and can verify this patch resolved the issue.


sethmuss (approved):

shetzel:
@David-Polehonski - Thank you for the PR!  I'll get this merged in but we should have a unit test for it.  Should be pretty simple.  I'd just copy the `'should create the package version create request with branch'` test (lines 319-337 of packageVersionCreate.test.ts) and modify it to pass in a `dependency.branch` and verify the correct branch was passed to the `packageCreateStub`.

shetzel (approved):

avesolovksyy:
@David-Polehonski @shetzel @cristiand391 
We faced with the same issue last week and suggested a bit different fix here: https://github.com/forcedotcom/cli/issues/2183

Our suggestion:
`const branch = (dependency.branch || dependency.branch === '') ? dependency.branch : this.options.branch;`

Current fix by @David-Polehonski :
`const branch = dependency.branch === '' ? this.options.branch : dependency.branch;`

Our concern is that current fix introduce a wrong behaviour if branch is specified for dependency explicitly as empty, e.g.:
```
"dependencies": [
  {
    "package": "My Base Package",
    "versionNumber": "1.8.0.LATEST",
    "branch": ""
  }
]
```
In such case, if we create a package version from command line specifying branch key explicitly (e.g. `--branch=myTestBranch`), it should still respect empty branch dependency (according to sfdx guide), but with @David-Polehonski fix seems like it's going to be overwritten and would be looking for 'My Base Package' version dependency within myTestBranch, which is incorrect.

PR #319, opened 2023-06-09
fix: added unit tests to better describe branch version selection behaviour and corrected issues

David-Polehonski:
As per @avesolovksyy feedback over [here](https://github.com/forcedotcom/cli/issues/2183) I have added 2 additional unit tests which I believe better describe the intended behaviour and amended the branch checking condition to resolve the regressions.

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -166,7 +166,7 @@ export class PackageVersionCreate {
     const buildNumber = versionNumber.build;
 
     // use the dependency.branch if present otherwise use the branch of the version being created
-    const branch = dependency.branch === '' ? this.options.branch : dependency.branch;
+    const branch = dependency.branch === undefined ? this.options.branch : dependency.branch;
     const branchString = !branch || branch === '' ? 'null' : `'${branch}'`;
 
     // resolve a build number keyword to an actual number, if needed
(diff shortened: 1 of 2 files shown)


avesolovksyy (commented):
@David-Polehonski 
Thanks for the change to address my concern.
Looks good now as for me, i.e. it covers now all cases I'm aware of.

avesolovksyy (approved):

shetzel (approved):

src/package/packageVersionCreate.ts
```diff
@@ -166,7 +166,7 @@ export class PackageVersionCreate {
     const buildNumber = versionNumber.build;
 
     // use the dependency.branch if present otherwise use the branch of the version being created
-    const branch = dependency.branch === '' ? this.options.branch : dependency.branch;
+    const branch = dependency.branch === undefined ? this.options.branch : dependency.branch;
```
shetzel, line 169: @David-Polehonski - let's change lines 169-170 to this equivalent and less verbose code:

```javascript
const branch = dependency.branch ?? this.options.branch;
const branchString = branch ? `'${branch}'` : 'null';
```

David-Polehonski: @shetzel I've made the suggested modification; I agree it's less verbose but it's not quite equivalent as the ?? operator will treat `null` and `undefined` in the same way. I'm not certain what the expected behaviour would be if you specify your dependency with `branch: null` in the project definition.

SagarBijjam (approved):

src/package/packageVersionCreate.ts
shetzel: A packaging team member has confirmed that `null` and `undefined` should be equivalent.  I'll merge when tests are passing.  Thanks for the contributions @David-Polehonski and @avesolovksyy

PR #330, opened 2023-06-23
feat: support querying the Package2Version SObject from PackageVersion

shetzel:
Adds a static method to the `PackageVersion` class for querying the `Package2Version` SObject.

@W-13650725@

### DIFF
--- a/src/package/packageVersion.ts
+++ b/src/package/packageVersion.ts
@@ -40,7 +40,7 @@ type Package2Version = PackagingSObjects.Package2Version;
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/packaging', 'package_version');
 
-export const Package2VersionFields = [
+export const Package2VersionFields: Array<keyof Package2Version> = [
   'Id',
   'IsDeleted',
   'CreatedDate',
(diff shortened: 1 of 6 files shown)


src/package/packageVersion.ts
```diff
@@ -336,6 +343,43 @@ export class PackageVersion {
...
+  ): Promise<T[]> {
+    let query = `SELECT ${fields.toString()} FROM Package2Version`;
+    if (whereClause) {
+      query += ` ${whereClause}`;
+      if (whereClauseItems) {
+        query += ' LIMIT 2000';
+        return queryWithInConditionChunking<T>(query, whereClauseItems, '%IDS%', connection);
+      }
+    }
+    query += ' LIMIT 2000';
```
mshanemc, line 375: when using SOQL `limit`: is there a `order by` that makes it more likely the user gets relevant results?
ex: `order by lastModifiedDate DESC`?

mshanemc (approved):
approved with a suggestion

test/utils/packageUtils.test.ts
```diff
@@ -82,6 +82,28 @@ describe('packageUtils', () => {
       expect(queryStub.callCount).to.equal(2);
     });
   });
+  describe('getPackageVersionNumber', () => {
+    it('should return build numbers when includeBuild=true', async () => {
```
mshanemc, line 86: these should not be async

src/package/packageVersion.ts
```diff
@@ -336,6 +343,47 @@ export class PackageVersion {
...
+   * @param fields The fields to include in the returned data. Defaults to all fields.
+   * @param whereClause Optional where clause to filter the query.
+   * @param whereClauseItems Optional array of where clause items to match. The query is chunked,
+   * meaning broken into multiple queries when the query length would exceed the maximum char limit.
+   * When defining items here, the `whereClause` argument must use this token for the item
+   * replacement: `'%IDS%'`.
+   * @param orderBy Optional order by clause. Defaults to LastModifiedDate descending.
+   * @returns Results from querying the Package2Version SObject.
+   */
+  public static async queryPackage2Version<T extends Record<string, unknown> = Record<string, unknown>>(
```
mshanemc, line 363: why is this generic?  isn't the type for Package2Version concrete?

mshanemc: also, if this is a brand new API, I think destructuring the params to an typed object is nicer since there are a bunch of optional ones

shetzel: It's generic because the return shape doesn't necessarily match `Package2Version` and could be a subset of it.  I really just wanted a way for the caller to define the type and not just be `Partial<Package2Version>`.  There's probably a better way though.

As for the params, it did grow from what I originally expected so an options object makes sense.  I can change that.

PR #492, opened 2024-01-13
feat: additional fields for package version result (VersionNumber, CodeCoverage)

Rocko1204:
### Helpful additional fields for PackageVersionCreate result

After implementing the last pull request #486 (thx for that), it turned out that two more fields were needed in the result.

| SObject | Field | Type | Example
|-----:|-----------|---------|----------|
|     Package2Version| CodeCoverage|number|78
|     Package2Version| VersionNumber|string|1.2.1.5


shetzel:
@Rocko1204 - thanks for the contribution and sorry for the delay.  This is under review and we'll work with you to get it in.  There can be performance issues around CodeCoverage so I'd like to ensure no perf degradation with these changes.  I believe this should be ok in this case though.

t7job (approved):

sethmuss (approved):

shetzel (approved):

PR #499, opened 2024-02-01
fix: set string replace env var before convert

shetzel:
Sets `SF_APPLY_REPLACEMENTS_ON_CONVERT=true` before converting packaged metadata so that strings are replaced automatically without the need for setting this env var explicitly.  String replacement happens automatically during a deploy, so this matches the expected behavior.  After conversion, the env var is set back to the original state.

@W-14914086@


shetzel:
Decided to do this in the plugin instead

PR #500, opened 2024-02-02
fix: allow increased query fetch limits

shetzel:
Updated the code to fetch 10,000 records max by default.  To set it higher than that use the `SF_ORG_MAX_QUERY_LIMIT` env var.

@W-13047283@

### DIFF
--- a/src/package/package.ts
+++ b/src/package/package.ts
@@ -5,6 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { Connection, Messages, SfError, SfProject } from '@salesforce/core';
+import { env } from '@salesforce/kit';
 import {
   ConvertPackageOptions,
   PackageCreateOptions,
@@ -118,12 +119,13 @@ export class Package {
    * @param connection
    */
   public static async list(connection: Connection): Promise<PackagingSObjects.Package2[]> {
+    const maxFetch = env.getNumber('SF_ORG_MAX_QUERY_LIMIT') ?? 10_000;
     return (
       await connection.tooling.query<PackagingSObjects.Package2>(
         `select ${this.getPackage2Fields(connection).toString()} from Package2 ORDER BY NamespacePrefix, Name`,
         {
           autoFetch: true,
-          maxFetch: 10_000,
+          maxFetch,
         }
       )
     )?.records;
--- a/src/package/packageVersionList.ts
+++ b/src/package/packageVersionList.ts
@@ -6,6 +6,7 @@
  */
 
 import { Connection, Logger, Messages } from '@salesforce/core';
+import { env } from '@salesforce/kit';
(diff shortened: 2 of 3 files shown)


src/package1/package1Version.ts
```diff
@@ -125,8 +125,14 @@ export class Package1Version implements IPackageVersion1GP {
     const query = `SELECT Id,MetadataPackageId,Name,ReleaseState,MajorVersion,MinorVersion,PatchVersion,BuildNumber FROM MetadataPackageVersion ${
       id ? `WHERE MetadataPackageId = '${id}'` : ''
     } ORDER BY MetadataPackageId, MajorVersion, MinorVersion, PatchVersion, BuildNumber`;
+    const maxFetch = env.getNumber('SF_ORG_MAX_QUERY_LIMIT') ?? 10_000;
```
mshanemc, line 128: it's not necessary to pass in the maxFetch since sfdx-core do exactly what line 128 is doing automatically https://github.com/forcedotcom/sfdx-core/blob/e7c496ef7b004037b41862c46173de9d27608ad5/src/org/connection.ts#L370-L371

I think what we were really missing was just `autoFetch:true`

src/package/packageVersionList.ts
```diff
@@ -63,10 +64,12 @@ export async function listPackageVersions(
   connection: Connection,
   options?: PackageVersionListOptions
 ): Promise<QueryResult<PackageVersionListResult>> {
+  const maxFetch = env.getNumber('SF_ORG_MAX_QUERY_LIMIT') ?? 10_000;
```
mshanemc, line 67: see below comment

src/package/packageVersionList.ts
```diff
@@ -6,6 +6,7 @@
  */
 
 import { Connection, Logger, Messages } from '@salesforce/core';
+import { env } from '@salesforce/kit';
```
mshanemc, line 9: remove, see below

src/package/package.ts
```diff
@@ -5,6 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { Connection, Messages, SfError, SfProject } from '@salesforce/core';
+import { env } from '@salesforce/kit';
```
mshanemc, line 8: remove, see below

mshanemc (changes requested):

mshanemc (approved):

mshanemc:
QA:

ran `sf package version list` against na40 hub with linked library.

✅  Saw 10k records, got a warning `Warning: The query result is missing 1064 records due to a 10000 record limit. Increase the number of records returned by setting the config value "maxQueryLimit" or the environment variable "SF_ORG_MAX_QUERY_LIMIT" to 11064 or greater than 10000.`
✅ error goes to stderr in both human and json

PR #535, opened 2024-03-28
@W-14742535@ Fix: definition-file keys case sensitivity

mradulsf:
The creation of the scratch org (build org) during the “package version create” command results in the org created without snapshot, because the snapshot property is null, even though the snapshot is specified in the scratch org definition file as "snapShot". 

While the scratch org is successfully created using the ‘snapShot’ property (uppercase ‘S’) with the “scratch org create” command. We should match this behavior.

- Actual Result

Scratch org (build org) created without the properties of Snapshot. For ex: Profiles that are created on the snapshotted org are missing in the build org.

- Expected Result

Scratch org (build org) should be created with the properties of Snapshot. The "package version create" command should not fail due to a missing custom profile that was from the snapshotted org.

@W-14742535@

### DIFF
--- a/src/utils/packageUtils.ts
+++ b/src/utils/packageUtils.ts
@@ -510,7 +510,10 @@ export function copyDescriptorProperties(
     packageDescriptorJsonCopy,
     Object.fromEntries(
       ['country', 'edition', 'language', 'features', 'orgPreferences', 'snapshot', 'release', 'sourceOrg'].map(
-        (prop) => [[prop], definitionFileJsonCopy[prop]]
+        (prop) => {
+          const matchCase = Object.keys(definitionFileJsonCopy).find((key) => key.toLowerCase() === prop.toLowerCase());
+          return [[prop], matchCase ? definitionFileJsonCopy[matchCase] : undefined];
+        }
       )
     )
   ) as PackageDescriptorJson;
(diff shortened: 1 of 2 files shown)


mradulsf:
test

mradulsf:
retest

mradulsf:
retest

src/utils/packageUtils.ts
```diff
@@ -510,7 +517,10 @@ export function copyDescriptorProperties(
     packageDescriptorJsonCopy,
     Object.fromEntries(
       ['country', 'edition', 'language', 'features', 'orgPreferences', 'snapshot', 'release', 'sourceOrg'].map(
-        (prop) => [[prop], definitionFileJsonCopy[prop]]
+        (prop) => {
+          const matchCase = Object.keys(definitionFileJsonCopy).find((key) => key.toLowerCase() === prop.toLowerCase());
+          return [[prop], matchCase ? definitionFileJsonCopy[matchCase] : undefined];
+        }
```
shetzel, line 523: This should be equivalent code as line 522 since an object lookup that doesn't resolve to anything is `undefined`, including if `matchCase` is `undefined`.
`return [[prop], definitionFileJsonCopy[matchCase]];`

mradulsf: @shetzel doing that change throws an error - `Type 'undefined' cannot be used as an index type.ts(2538)`

Another option is to set the default of `undefined`
`return [[prop], definitionFileJsonCopy[matchCase ?? 'undefined']];`

shetzel: Ok that works then.  Thanks!

PR #548, opened 2024-04-16
fix: move verbose field to default fields

mradulsf:
Move the non-CodeCoverage fields from the verbose query to the default query so that JSON responses are more accurate.

@W-15398724@

### DIFF
--- a/src/package/packageVersionList.ts
+++ b/src/package/packageVersionList.ts
@@ -71,7 +71,7 @@ export async function listPackageVersions(
   return connection.autoFetchQuery<PackageVersionListResult & Schema>(query, { tooling: true });
 }
 
-function constructQuery(connectionVersion: number, options?: PackageVersionListOptions): string {
+export function constructQuery(connectionVersion: number, options?: PackageVersionListOptions): string {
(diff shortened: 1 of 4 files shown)


SagarBijjam (commented):
@mradulsf  the WI mentioned about similarly affected packaging commands. have you come across any. Like `package:version:create:report` command that have similar fields

mradulsf:
@SagarBijjam Yes, I made change to another file for package version report. Is that not the one you are talking about?

shetzel (changes requested):
You need to update `packageTest.nut.ts` line 279 to now expect the `Id` as a property of the returned result since you added it to the query.
In other words, it should now be:
`expect(result).to.have.property('Id');`

Our preference is to test public APIs rather than exposing a private function and unit testing that.

SagarBijjam:
> @SagarBijjam Yes, I made change to another file for package version report. Is that not the one you are talking about?

I was referring to `package:version:create:report`, but it seems that the command doesn't have similar verbose fields. So skip it

shetzel (approved):

SagarBijjam (approved):

PR #549, opened 2024-04-18
test: add xNUTs to GHA

shetzel:
Adds the plugin-packaging NUTs to the test workflow for GHA so we can know if changes to this library will break the plugin.

@W-15534460@

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,6 +36,6 @@ jobs:
       externalProjectGitUrl: 'https://github.com/salesforcecli/plugin-packaging'
       command: 'yarn test:nuts:package'
       os: ${{ matrix.os }}
-      preExternalBuildCommands: 'yarn upgrade @salesforce/core; npx yarn-deduplicate; yarn install'
+      preExternalBuildCommands: 'shx rm -rf node_modules/@salesforce/core'
       useCache: false
     secrets: inherit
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -24,3 +24,19 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'windows-latest']
+    with:
+      packageName: '@salesforce/packaging'
+      externalProjectGitUrl: 'https://github.com/salesforcecli/plugin-packaging'
+      command: 'yarn test:nuts:package'
+      os: ${{ matrix.os }}
+      preSwapCommands: 'yarn upgrade @salesforce/core; npx yarn-deduplicate; yarn install'
+      preExternalBuildCommands: 'npm why @salesforce/core --json'
```
WillieRuemmele, line 40: do we need this, or was it for debugging?

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---
:x: : I created a PR off of this branch https://github.com/forcedotcom/packaging/actions/runs/8785920088/job/24107628626?pr=555 and in the actions it failed to build

.github/workflows/test.yml
shetzel: Arguably both.  I added it for debugging but it won't harm anything to keep it.  That way when tsc fails when building the xNUT repo we can see why, at least for `@salesforce/core`.  I can remove it though.

PR #563, opened 2024-05-09
fix: add a method that returns a 04t from a partial package descriptor entry

shetzel:
Adds a public static method to the `SubscriberPackageVersion` API that returns the SubscriberPackageVersionId (04t) given a partial packageDirectories entry (`versionNumber` and `packageId` fields, as well as an optional `branch` field).

https://github.com/forcedotcom/cli/discussions/2721
@W-15611712@


src/package/packageVersionCreate.ts
```diff
@@ -91,71 +91,15 @@ export class PackageVersionCreate {
...
+   * A dependency in the project config file may be specified using either a subscriber package version id (04t)
    * or a package Id (0Ho) + a version number.  Additionally, a build number may be the actual build number, or a
    * keyword: LATEST or RELEASED (meaning the latest or released build number for a given major.minor.patch).
    *
    * This method resolves a package Id + version number to a subscriber package version id (04t)
    * and adds it as a SubscriberPackageVersionId parameter in the dependency object.
    */
-  private async retrieveSubscriberPackageVersionId(dependency: PackageDescriptorJson): Promise<PackageDescriptorJson> {
-    await this.validateDependencyValues(dependency);
-    if (dependency.subscriberPackageVersionId) {
```
WillieRuemmele, line 152: do we need to call `resolveSubscriberPackageVersionId` here now since we moved this into its own method?

WillieRuemmele (approved):

PR #590, opened 2024-06-11
refactor: type consolidation

mshanemc:
uses types for sfProject, PackageDir etc from sfdx-core
refactor some un-bundleable geLogger stuff
othr basic TS cleanup (remove assertions, simplify types)

### DIFF
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -330,7 +330,7 @@ export class PackageVersionCreate {
     const mdOptions: MDFolderForArtifactOptions = {
       deploydir: packageVersMetadataFolder,
       sourceDir: sourceBaseDir,
-      sourceApiVersion: (this.project?.getSfProjectJson()?.get('sourceApiVersion') as string) ?? undefined,
+      sourceApiVersion: this.project?.getSfProjectJson()?.get('sourceApiVersion'),
     };
 
     await fs.promises.mkdir(packageVersBlobDirectory, { recursive: true });
--- a/src/package/packageVersionCreate.ts
+++ b/src/package/packageVersionCreate.ts
@@ -14,7 +14,6 @@ import {
   Logger,
   Messages,
   NamedPackageDir,
-  PackageDir,
   ScratchOrgInfo,
   SfdcUrl,
   SfProject,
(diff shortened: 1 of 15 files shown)


test/package/ancestry.nut.ts
```diff
@@ -0,0 +1,170 @@
+/*
```
mshanemc, line 1: this is completely independent from the e2e nut so I split it out (can run parallel)

test/package/packageTest.nut.ts
```diff
@@ -128,6 +124,7 @@ describe('Integration tests for @salesforce/packaging library', () => {
       const projectFile = JSON.parse(dxPjsonData) as ProjectJson;
       expect(projectFile).to.have.property('packageDirectories').with.length(1);
       expect(projectFile.packageDirectories[0]).to.include.keys(['package', 'versionName', 'versionNumber']);
+      assert(isPackagingDirectory(projectFile.packageDirectories[0]));
       expect(projectFile.packageDirectories[0].package).to.equal(pkgName);
```
mshanemc, line 128: these all presume packageDirectories[0] is the flavor that has the full packaging info.  asserting that via the typeguard makes everything downstream of that assertion type-aware

src/package/packageVersionCreate.ts
```diff
@@ -649,16 +650,19 @@ export class PackageVersionCreate {
         if (!packageName) throw messages.createError('errorMissingPackage', [this.options.packageId]);
       }
       packageObject = this.project.findPackage(
-        (namedPackageDir) => namedPackageDir.package === packageName || namedPackageDir.name === packageName
+        (namedPackageDir) =>
+          (isPackagingDirectory(namedPackageDir) && namedPackageDir.package === packageName) ||
```
WillieRuemmele, line 654: I don't think this logic matches, should it be 
`isPackagingDirectory(namedPackageDir) &&(namedPackageDir.package === packageName || namedPackageDir.name === packageName)`?

src/package1/package1Version.ts
```diff
@@ -172,3 +136,32 @@ export class Package1Version implements IPackageVersion1GP {
     return (await this.connection.tooling.query<PackagingSObjects.MetadataPackageVersion>(query)).records;
   }
 }
+
+const packageUploadPolling = async (
```
WillieRuemmele, line 140: I believe this is only applicable to 1GP packages, and why it was kept under the `Package1Version` class.  I think keeping that clarity to the consumer is good

WillieRuemmele (changes requested):

src/package1/package1Version.ts
mshanemc: I'm not following.  it's not exported, so there shouldn't be any consumers outside the package1Version module, right?

WillieRuemmele (approved):

WillieRuemmele (approved):

PR #606, opened 2024-06-21
feat: display validated-async in version list and report

mradulsf:
Follow up from - https://github.com/forcedotcom/packaging/pull/591/commits

### DIFF
--- a/src/package/packageVersionList.ts
+++ b/src/package/packageVersionList.ts
@@ -74,8 +74,8 @@ export async function listPackageVersions(
 export function constructQuery(connectionVersion: number, options?: PackageVersionListOptions): string {
   // construct custom WHERE clause, if applicable
   const where = constructWhere(options);
-
-  let queryFields = defaultFields;
+  // Ensure we only include the async validation property for api version of v60.0 or higher.
+  let queryFields = connectionVersion > 60 ? [...defaultFields, 'ValidatedAsync'] : defaultFields;
   if (options?.verbose) {
     queryFields = [...queryFields, ...verboseFields];
     if (connectionVersion >= 57) {
--- a/src/package/packageVersionList.ts
+++ b/src/package/packageVersionList.ts
@@ -46,6 +46,9 @@ const verboseFields = ['CodeCoverage', 'HasPassedCodeCoverageCheck'];
 
 const verbose57Fields = ['Language'];
 
+// Ensure we only include the async validation property for api version of v60.0 or higher.
+const default61Fields = ['ValidatedAsync'];
+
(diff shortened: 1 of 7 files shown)


src/package/packageVersionReport.ts
```diff
@@ -39,6 +50,22 @@ const getLogger = (): Logger => {
   return logger;
 };
 
+function constructQuery(connectionVersion: number, verbose: boolean): string {
+  // Ensure we only include the async validation property for api version of v60.0 or higher.
+  let queryFields = connectionVersion > 60 ? [...defaultFields, 'ValidatedAsync'] : defaultFields;
+  verboseFields = connectionVersion > 60 ? [...verboseFields, 'EndToEndBuildDurationInSeconds'] : verboseFields;
```
azenoni, line 56: I think you need to include ValidatedAsync as well as EndToEndBuildDurationInSeconds for verbose fields

test/package/packageVersionList.test.ts
azenoni, line 1: Is there a similar test for version report that we need to update with the new fields?

src/package/packageVersionList.ts
```diff
@@ -74,8 +74,8 @@ export async function listPackageVersions(
 export function constructQuery(connectionVersion: number, options?: PackageVersionListOptions): string {
   // construct custom WHERE clause, if applicable
   const where = constructWhere(options);
-
-  let queryFields = defaultFields;
+  // Ensure we only include the async validation property for api version of v60.0 or higher.
+  let queryFields = connectionVersion > 60 ? [...defaultFields, 'ValidatedAsync'] : defaultFields;
```
azenoni, line 78: Can we follow the same pattern as the Language field? (ie have another array default60Fields). Just to keep it consistent

src/package/packageVersionReport.ts
mradulsf: From what I remember Dileep mentioned `ValidatedAsync` will have same behavior as `Validation Skipped` and the latter one is non-verbose field.

test/package/packageVersionList.test.ts
mradulsf: There are no tests for version report as of now.

src/package/packageVersionReport.ts
azenoni: Misread the code, looks okay.

test/package/packageVersionList.test.ts
shetzel: https://github.com/forcedotcom/packaging/blob/main/test/package/packageTest.nut.ts#L291

We prefer to test the public API both with unit tests and NUTs since the tests are more reliable and require less maintenance versus targeted unit tests of specific functions.

shetzel (approved):

PR #638, opened 2024-08-12
feat: make `SfProject` optional

cristiand391:
This PR updates some packaging classes to make `SfProject` optional.

See: https://github.com/salesforcecli/plugin-packaging/pull/771

Most of these classes were taking an `SfProject` instance along with a possible package ID **or** alias (defined in sfdx-project.json). By making `SfProject` optional, consumers of the library can call it without having to create a dummy sfdx-project.

Note for review:
there are classes/methods that do need a project b/c they write stuff on it after a package operation, most of these aren't affected by this change (`PackageVersion.updateProjectWithPackageVersion` was updated to throw if there's no project).

QA:

Check the plugin-packaging PR linked above, with this branch linked (or the prerelease) you should be able to use the listed commands outside a project.

@W-16286213@

### DIFF
--- a/src/package/package.ts
+++ b/src/package/package.ts
@@ -83,8 +83,9 @@ export class Package {
   public constructor(private options: PackageOptions) {
     let packageId = this.options.packageAliasOrId;
     if (!packageId.startsWith(packagePrefixes.PackageId)) {
-      packageId =
-        this.options.project.getPackageIdFromAlias(this.options.packageAliasOrId) ?? this.options.packageAliasOrId;
+      packageId = this.options.project
+        ? this.options.project.getPackageIdFromAlias(this.options.packageAliasOrId) ?? this.options.packageAliasOrId
+        : this.options.packageAliasOrId;
       if (packageId === this.options.packageAliasOrId) {
         throw messages.createError('packageAliasNotFound', [this.options.packageAliasOrId]);
       }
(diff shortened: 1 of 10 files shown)


src/package/package.ts
```diff
@@ -83,8 +83,9 @@ export class Package {
   public constructor(private options: PackageOptions) {
     let packageId = this.options.packageAliasOrId;
     if (!packageId.startsWith(packagePrefixes.PackageId)) {
-      packageId =
-        this.options.project.getPackageIdFromAlias(this.options.packageAliasOrId) ?? this.options.packageAliasOrId;
+      packageId = this.options.project
+        ? this.options.project.getPackageIdFromAlias(this.options.packageAliasOrId) ?? this.options.packageAliasOrId
+        : this.options.packageAliasOrId;
       if (packageId === this.options.packageAliasOrId) {
         throw messages.createError('packageAliasNotFound', [this.options.packageAliasOrId]);
```
cristiand391, line 90: the constructor will fail if you pass an alias without project or an invalid alias with a project instance.

src/package/package.ts
```diff
@@ -136,12 +137,12 @@ export class Package {
...
   public static async listVersions(
     connection: Connection,
-    project: SfProject,
+    project?: SfProject,
     options?: PackageVersionListOptions
   ): Promise<PackageVersionListResult[]> {
     // resolve/verify packages
     const packages = options?.packages?.map((pkg) => {
-      const id = project.getPackageIdFromAlias(pkg) ?? pkg;
+      const id = project ? project.getPackageIdFromAlias(pkg) ?? pkg : pkg;
```
cristiand391, line 145: these IDs are validated below

mshanemc (approved):

mshanemc (approved):

mshanemc:
QA for that last commit: from the plugin using the prerel, 

✅  ran with an alias and got the good helpful error message about the missing project
✅ ran with a valid Id and it worked fine

PR #690, opened 2024-12-12
feat: Create new package push upgrade class

btrn11:


src/package/package.ts
```diff
@@ -303,8 +303,8 @@ export class Package {
    *
    * @param force force a refresh of the package data
    */
-  public async getPackageData(force = false): Promise<PackagingSObjects.Package2 | undefined> {
-    if (!this.packageData ?? force) {
+  public async getPackageData(): Promise<PackagingSObjects.Package2 | undefined> {
```
tshopshireSalesforce, line 306: Let's not change this method since we don't own it. I'd make the change necessary for this to compile on your local but then not include it in the PR. If it gives us issues when attempting to submit we will have to fix it. Also, if you reverse the order in the if it should stop complaining and you it won't change the functionality

btrn11: Yes, this gave me issue when I tried to push my code, let me check again if we need to include this in PR.

btrn11: When I flipped the order, a lot of tests failed. And keeping it as it's would not allow me to push code.
Also I think removing force is similar since it's always false and those calling this method is not changing force param (always calling getPackageData() without param)

tshopshireSalesforce (approved):

src/package/packagePushUpgrade.ts
```diff
@@ -0,0 +1,66 @@
...
+  const where: string[] = [];
+
+  if (options?.packageId) {
+    where.push(`MetadataPackageVersion.MetadataPackage = '${options.packageId}'`);
+  }
+  return where.length > 0 ? `WHERE ${where.join(' AND ')}` : '';
+}
+
+function getListQuery(): string {
+  const QUERY = 'SELECT Id, PackageVersion, Status' + 'FROM PackagePushRequest ' + '%s'; // WHERE, if applicable
```
WillieRuemmele, line 64: NIT: can make this inline... `return 'SELECT Id...`

PR #710, opened 2025-01-15
fix: use rest api for initiating push upgrade request and bulk api v2 for push upgrade jobs

btrn11:


src/package/packagePushUpgrade.ts
```diff
@@ -130,36 +136,40 @@ export class PackagePushUpgrade {
...
-        throw new Error('Failed to create PackagePushRequest');
-      }
+      const pushRequestResult: PackagePushRequestResult = await connection.request({
+        method: 'POST',
+        url: '/services/data/v55.0/sobjects/packagepushrequest/',
+        body: JSON.stringify(packagePushRequestBody),
+      });
 
       // Create PackagePushJob for each org using Bulk API v2
       const job = connection.bulk2.createJob({ object: 'PackagePushJob', operation: 'insert' });
```
btrn11, line 151: @cristiand391 Continuing this [conversation](https://github.com/forcedotcom/packaging/pull/703#discussion_r1917358329) in this new PR
cc: @tshopshireSalesforce

cristiand391: > @cristiand391 Thanks, how can I check if PackagePushJob is a valid object or not for Bulk API (For example it's not an object in Tooling API but has a REST endpoint). Right now the program is stuck at createJob line (not hitting catch block). I want to make sure we can use Bulk API here or we would have to do a loop of different REST API calls.

I ran this bulk query with the CLI and didn't get any error so I assume the object supports bulk operations (at least query):

```
sf data export bulk -q 'select PackagePushRequestId from PackagePushJob' -o na40 --wait 10 --output-file res.csv
```

Screenshot 2025-01-15 at 6 40 19 PM

afaik there's no list of objects that state they support bulk so I just do a bulk query/import to test.

cristiand391: @btrn11 you can see more debug logs from jsforce by setting this env var:
```
JSFORCE_LOG_LEVEL=DEBUG
```

also, do you see the bulk job beign created in the org? setup -> `Bulk Data Load Jobs`

btrn11: @cristiand391 I saw the job has been created but it failed because of the below error. I checked and there are no other job with the same id, I also made sure to abort any ongoing jobs previously but still encounter it. Do you know what this error means and if I should do anything differently in my code. I followed the steps in jsforce docs. Thanks!
image

btrn11: But in the code, I only call job.uploadData once, why are there multiple PUT requests? Also do you know which team I should reach out to regarding bulk api? Thanks!

cristiand391: I don't see multiple PUT calls in your code/logs, was just mentioning what could cause the error.

You could ask owners of the bulk API in the `#discuss-enterprise-api` in our slack workspace.

src/package/packagePushUpgrade.ts
```diff
@@ -130,36 +136,46 @@ export class PackagePushUpgrade {
...
-      if (!pushRequest.success) {
-        throw new Error('Failed to create PackagePushRequest');
-      }
+      };
 
-      // Create PackagePushJob for each org using Bulk API v2
-      const job = connection.bulk2.createJob({ object: 'PackagePushJob', operation: 'insert' });
+      const pushRequestResult: PackagePushRequestResult = await connection.request({
+        method: 'POST',
+        url: '/services/data/v62.0/sobjects/packagepushrequest/',
```
btrn11, line 146: @cristiand391 Can this version (v62.0) be dynamic in some way?

cristiand391: yeah, you could do something like this to grab whatever API version is set on the `Connection` object:
```ts
url: `/services/data/v${connection.getApiVersion()}/sobjects/packagepushrequest/`,
```

even better:
use jsforce's `sobject.create` and let it handle requests for you 😉 
https://jsforce.github.io/document/#create

btrn11: Got it, thanks!

src/package/packagePushUpgrade.ts
```diff
@@ -185,13 +201,14 @@ export class PackagePushUpgrade {
...
-            `Job ${index + 1} Error:${JSON.stringify(job?.sf__Error, null, 2)}`
+            `Job ${index + 1}: Error:${JSON.stringify(job?.sf__Error, null, 2)}`
         )
         .join('');
 
       await fs.writeFile(outputFile, errorContent, 'utf-8');
+      throw new Error(`Push upgrade failed, job errors have been written to file: ${outputFile}`);
     } catch (error) {
-      throw new Error('Error when saving job errors to file' + (error as Error).message);
+      throw new Error('Error when saving job errors to file. ' + (error as Error).message);
```
shetzel, line 211: Prefer using `SfError` from `@salesforce/core`.  It has a static `wrap()` method that is pretty useful especially to preserve original error stack and cause.

Also, the error thrown on line 209 will be caught here and the error will be misleading/confusing.

src/package/packagePushUpgrade.ts
```diff
@@ -185,13 +201,14 @@ export class PackagePushUpgrade {
         .map(
           (job, index) =>
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            `Job ${index + 1} Error:${JSON.stringify(job?.sf__Error, null, 2)}`
+            `Job ${index + 1}: Error:${JSON.stringify(job?.sf__Error, null, 2)}`
```
shetzel, line 204: If `job.sf__Error` is ever undefined this will throw an error.  Might be better to not disable the lint rule and ensure it exists in code unless you are 100% sure it will always be defined at this point.

src/package/packagePushUpgrade.ts
```diff
@@ -130,36 +136,46 @@ export class PackagePushUpgrade {
...
+        url: `/services/data/v${connection.getApiVersion()}/sobjects/packagepushrequest/` + pushRequestResult?.id,
+        body: JSON.stringify({ Status: 'Pending' }),
+      });
+
       return {
-        PushRequestId: pushRequest.id,
+        PushRequestId: pushRequestResult.id,
         ScheduledStartTime: scheduleTime,
         Status: 'Pending',
       };
```
shetzel, line 181: There's a lot of stuff going on in the `try` that could cause `err instanceof Error` to be true.  Will `throw applyErrorAction(massageErrorMessage(err));` handle it all?

btrn11: I see, do you have any suggestions on this? Thanks!

btrn11:
@shetzel I also addressed this one too, can you please take a look? Thank you!

src/package/packagePushUpgrade.ts
```diff
@@ -128,38 +134,50 @@ export class PackagePushUpgrade {
...
+      await job.poll();
 
       // If there are any errors for a job, write all specific job errors to an output file
       const jobErrors = await job.getFailedResults();
 
       if (jobErrors.length > 0) {
-        await this.writeJobErrorsToFile(pushRequest.id, jobErrors);
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        const error: SfError = await this.writeJobErrorsToFile(pushRequestResult?.id, jobErrors);
+        return error;
```
cristiand391, line 170: you should avoid passing errors as values through functions unless, it's very easy for callers to ignore these.
suggestion:
make `writeJobErrorsToFile` return the filePath that was written and build & throw the error here, `sf` will catch the error and finish the process.

src/package/packagePushUpgrade.ts
```diff
@@ -128,38 +134,50 @@ export class PackagePushUpgrade {
...
 
-      await job.check();
+      // Create PackagePushJob for each org using Bulk API v2
+      const job = connection.bulk2.createJob({ operation: 'insert', object: 'PackagePushJob' });
+
+      await job.open();
+
       await job.uploadData(pushJobs);
       await job.close();
+      await job.poll();
```
cristiand391, line 162: make sure to increase the polling timeout/interval, the default values are:
https://github.com/jsforce/jsforce/blob/4959579c16c148bdf2b2b9efea6f73167d525fa0/src/api/bulk2.ts#L181C1-L194C1

If you expect these to vary depending on user payload maybe allow them to configure these values via flags?

`job.bulk`:
https://github.com/jsforce/jsforce/blob/4959579c16c148bdf2b2b9efea6f73167d525fa0/src/api/bulk2.ts#L764

src/package/packagePushUpgrade.ts
```diff
@@ -174,24 +192,21 @@ export class PackagePushUpgrade {
...
     try {
       await fs.mkdir(outputDir, { recursive: true });
 
       const errorContent = jobErrors
-        .map(
-          (job, index) =>
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            `Job ${index + 1} Error:${JSON.stringify(job?.sf__Error, null, 2)}`
-        )
+        .map((job, index) => `Job ${index + 1}: Error:${JSON.stringify(job?.sf__Error, null, 2)}`)
```
cristiand391, line 203: you can also get the failed results as a CSV and write it to a file:
https://github.com/jsforce/jsforce/blob/4959579c16c148bdf2b2b9efea6f73167d525fa0/src/api/bulk2.ts#L921

cristiand391 (changes requested):
left some suggestions and requested changes about funcs returning errors

src/package/packagePushUpgrade.ts
btrn11: @tshopshireSalesforce What do you think?

btrn11: Our jobs might take hours to run @cristiand391

src/package/packagePushUpgrade.ts
btrn11: I used that function here: https://github.com/forcedotcom/packaging/pull/710/files#diff-faa2c36b55b358292065b7eb26aa221eda7fba82f4a81f957f0bc5433a5b534dR165

src/package/packagePushUpgrade.ts
cristiand391: @btrn11 @tshopshireSalesforce `await job.poll()` will finish after the bulk job is completed.
If you want to remove the poll then you users should have a way to resume this operation to get results (bulk job failures). I've looked at your design doc and don't see how this is handled.

src/package/packagePushUpgrade.ts
cristiand391: yes, I meant that you can pass `true` to that function if you want to get the failures as a big CSV string.
In this code you are stringifyng it again so maybe the CSV works better for you.

cristiand391: related: https://github.com/forcedotcom/packaging/pull/710#discussion_r1926102679

src/package/packagePushUpgrade.ts
btrn11: > `await job.poll()` will finish after the bulk job is completed.

This should work right? This is what we want actually.

btrn11: Not sure what would be the best approach here, if increasing poll time, what numbers do you recommend? @cristiand391

btrn11: > if you expect this the bulk job to take hours then this should be an async operation

@cristiand391 Can you explain further what you mean on this? Thanks!

tshopshireSalesforce: This creation of entities should not take hours. It kicks off an async job that may take a long time but we don't need to poll that long. This is a lightweight entity to create.

cristiand391: @btrn11 @tshopshireSalesforce ok, just to be clear I'm talking about the Bulk V2 insert job:
`await job.poll` will resolve successfully once the job finished being processed or throw an error if a failure happened:
https://github.com/jsforce/jsforce/blob/4959579c16c148bdf2b2b9efea6f73167d525fa0/src/api/bulk2.ts#L799C1-L807C20

### Polling time
This depends on how much you think the job will take to finish (qty of records to insert). If this depends on the user payload you might need to expose it via a `--wait` flag. If no flag, then at least 5/10min (or more, depending on your experience running the command).

### Handling job failures
You handle insert failures here:
https://github.com/forcedotcom/packaging/blob/3e1d523a72fc86c2f43350255e9f7c755f06bec8/src/package/packagePushUpgrade.ts#L164

but what happens if `await job.poll` throws (job failure in the org)? do you need to rollback any change here? or do you expect the customer to resume the operation via a new command? does the customer need to debug the job failure?

tshopshireSalesforce (approved):

src/package/packagePushUpgrade.ts
btrn11: If the job.poll fails we will just rely on the catch block. The user would have to redo the commands, if there's any issue with the job itself after being created then we write those errors to the file.

cristiand391: ok, then you just need to decide what are you going to do with the poll timeout, either way (hardcode or via `--wait`) should be better than the 30s default.

btrn11: I see, thanks. We'll test this out, if it becomes an issue we'll revisit.

PR #712, opened 2025-01-28
@W-17409066 feat - push upgrade request abort cli support

mradulsf:


LICENSE.txt
tshopshireSalesforce, line 1: No need to modify this file.

src/package/packagePushUpgrade.ts
```diff
@@ -187,6 +189,48 @@ export class PackagePushUpgrade {
...
+
+      if (!queryResult.records || queryResult.records.length === 0) {
+        throw new Error(`No PackagePushRequest found with Id: ${options.packagePushRequestId}`);
+      }
+
+      const pushRequest = queryResult.records[0];
+      // Validate the current status
+      if (!['Created', 'Pending'].includes(pushRequest.Status)) {
+        throw new Error(
+          `Cannot abort PackagePushRequest with status '${pushRequest.Status}'. Abortion is only allowed for 'Created' or 'Pending' statuses.`
```
tshopshireSalesforce, line 209: Is this the error message Penelope provided? She may have already written one for this use case if you check her doc.

src/package/packagePushUpgrade.ts
```diff
@@ -187,6 +189,48 @@ export class PackagePushUpgrade {
...
+        );
+      }
+
+      // Abort the push request by setting its status to "Canceled"
+      await connection.request({
```
tshopshireSalesforce, line 214: Hmm, this feels strange to be making a manual request to this url. Is there a typescript API to use somewhere instead of manually crafting the url and payload? I'll take a look when I have some free time but I imagine this may be done elsewhere in this repo.

src/package/packagePushUpgrade.ts
```diff
@@ -187,6 +189,48 @@ export class PackagePushUpgrade {
...
+        method: 'PATCH',
+        url:
+          `/services/data/v${connection.getApiVersion()}/sobjects/packagepushrequest/` + options.packagePushRequestId,
+        body: JSON.stringify({ Status: 'Canceled' }),
+      });
+
+      // Return the updated PackagePushRequest details
+      return {
+        PushRequestId: pushRequest.Id,
+        Status: 'Canceled',
```
tshopshireSalesforce, line 224: I think it makes sense to have this method return true/false instead of this PackagePushUpgradeAbortResult. We never return a different status and the pushRequest id is the same as what was passed in so it isn't all that valuable of a returned value. I think true/false and then in packaging-plugin we can display another message if it wasn't successful.

src/package/packagePushUpgrade.ts
```diff
@@ -294,3 +338,14 @@ async function queryJobFailureReasons(
...
 }
+
+function getPushRequestStatusWhereClause(options: PackagePushRequestAbortQueryOptions): string {
+  const where: string[] = [];
+  where.push(`Id = '${options.packagePushRequestId}'`);
+  return `WHERE ${where.join(' AND ')}`;
+}
+
+function getPushRequestStatusQuery(): string {
+  const QUERY = 'SELECT PackageVersionId, Id, Status FROM PackagePushRequest ' + '%s';
```
btrn11, line 349: I don't think you need PackageVersionId here.

LICENSE.txt
mradulsf: It was automatically done by lint I guess. want me to revert it?

src/package/packagePushUpgrade.ts
mradulsf: Make sense.

tshopshireSalesforce (approved):

PR #718, opened 2025-02-11
fix: Increase bulk time and delete bulk job that errors due to timeout

btrn11:


src/package/packagePushUpgrade.ts
```diff
@@ -182,10 +184,11 @@ export class PackagePushUpgrade {
         Status: 'Pending',
       };
     } catch (err) {
-      if (err instanceof Error) {
-        throw applyErrorAction(massageErrorMessage(err));
+      if (job && (err as Error).name === 'JobPollingTimeoutError') {
```
btrn11, line 187: @cristiand391 I saw your example but I figured we would want to delete the job **when** it's timeout error so I changed !== to ===. This way I assume the bulk job will be cancelled, therefore this will cancel our push upgrade job as well.

src/package/packagePushUpgrade.ts
```diff
@@ -154,13 +156,13 @@ export class PackagePushUpgrade {
...
       // Create PackagePushJob for each org using Bulk API v2
-      const job = connection.bulk2.createJob({ operation: 'insert', object: 'PackagePushJob' });
+      job = connection.bulk2.createJob({ operation: 'insert', object: 'PackagePushJob' });
 
       await job.open();
 
       await job.uploadData(pushJobs);
       await job.close();
-      await job.poll();
+      await job.poll(1000, 60_000);
```
btrn11, line 165: Increased bulk time to 1 minute

src/package/packagePushUpgrade.ts
cristiand391: note that `JobPollingTimeoutError` is a client-side error (from jsforce) because the library ran out of time waiting for the job to be finished, in the org it keeps being processed.

If you delete it like in L189 it will not rollback inserted records.

btrn11: I see, so we have to call the api again to abort/delete the push upgrade job ourselves here correct? Or does bulk api has a way to handle this?

src/package/packagePushUpgrade.ts
```diff
@@ -182,10 +184,11 @@ export class PackagePushUpgrade {
         Status: 'Pending',
       };
     } catch (err) {
-      if (err instanceof Error) {
-        throw applyErrorAction(massageErrorMessage(err));
+      if (job && (err as Error).name !== 'JobPollingTimeoutError') {
```
btrn11, line 187: @cristiand391 updated to !==

src/package/packagePushUpgrade.ts
```diff
@@ -154,13 +156,13 @@ export class PackagePushUpgrade {
...
       // Create PackagePushJob for each org using Bulk API v2
-      const job = connection.bulk2.createJob({ operation: 'insert', object: 'PackagePushJob' });
+      job = connection.bulk2.createJob({ operation: 'insert', object: 'PackagePushJob' });
 
       await job.open();
 
       await job.uploadData(pushJobs);
       await job.close();
-      await job.poll();
+      await job.poll(1000, 600_000);
```
btrn11, line 165: Increased time out to 10 mins

cristiand391 (approved):

PR #745, opened 2025-06-10
feat: added bundle create

dawson-david-salesforce:
added the bundle create command and the bases for other commands in the same vein, e.g the bundle Utils. Heavily based of the package implementation. Relies upon a modified sfdx-core repo.

### DIFF
--- a/src/interfaces/bundleInterfacesAndType.ts
+++ b/src/interfaces/bundleInterfacesAndType.ts
@@ -4,11 +4,10 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-export {
-  INSTALL_URL_BASE,
-  getContainerOptions,
-  getPackageVersionStrings,
-  getPackageVersionNumber,
-} from './packageUtils';
 
-export { massageErrorMessage } from './bundleUtils';
+// import { BundleSObjects } from './bundleSObjects';
+
+export type BundleCreateOptions = {
+  BundleName: string;
+  Description: string;
+};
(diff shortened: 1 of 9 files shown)


src/interfaces/bundleInterfacesAndType.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
```
btrn11, line 2: Change to 2025 for all files

src/package/packageBundle.ts
```diff
@@ -0,0 +1,28 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { Connection, SfProject } from '@salesforce/core';
+import { BundleCreateOptions } from '../interfaces';
+import { createBundle } from './packageBundleCreate';
+
+export class PackageBundle {
+  /**
+   * Create a new package.
```
btrn11, line 14: Change to create a new bundle

src/package/packageBundleCreate.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+import { Connection, SfError, SfProject } from '@salesforce/core';
+import { BundleEntry } from '@salesforce/schemas/src/sfdx-project/bundleEntry';
+import { BundleSObjects, BundleCreateOptions } from '../interfaces';
+import { massageErrorMessage } from '../utils/bundleUtils';
+
+type Bundle2Request = Pick<BundleSObjects.Bundle, 'BundleName' | 'Description'>;
+
+// This is cargo cult programing, this function does nothing, but is here incase I need to add context to the request with the project
+export function createBundleRequestFromContext(project: SfProject, options: BundleCreateOptions): Bundle2Request {
```
btrn11, line 15: What do you mean by adding context here? It still only returns the Bundle2Request type correct for the createBundle function to use?

btrn11: If so can you rename this method, or determine again if we need it or not?

PR #747, opened 2025-06-11
W-18760756: Adding code coverage option to the convert command

smotwani11:
This PR adds a code coverage [-c] option to the convert command. This option when specified will allow the conversion command to calculate and store the code coverage percentage by running the packaged Apex tests included in this package version.

@W-18760756@


src/package/packageConvert.ts
```diff
@@ -254,6 +256,7 @@ async function createRequestObject(
     InstallKey: options.installationkey,
     Instance: options.buildinstance,
     IsConversionRequest: true,
+    CalculateCodeCoverage: options.codecoverage ?? false,
```
ganeshnt, line 259: didn't this field/option already existed on/for native package version??

ganeshnt (approved):

src/package/packageConvert.ts
smotwani11: Yes, it did. It just wasn't being passed in convert command.

PR #751, opened 2025-06-18
feat: created bundle version create

dawson-david-salesforce:
created bundle version create, and report/ list feature for requests


LICENSE.txt
tshopshireSalesforce, line 1: Please remove this file from the PR.

src/package/packageBundleVersionCreate.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+  }
+
+  public static async getCreateStatuses(
+    connection: Connection,
+    status?: BundleSObjects.PkgBundleVersionCreateReqStatus,
+    createdLastDays?: number
+  ): Promise<BundleSObjects.PackageBundleVersionCreateRequestResult[]> {
+    let query =
+      'SELECT Id, RequestStatus, PackageBundle.Id, PackageBundle.BundleName, PackageBundleVersion.Id, ' +
```
tshopshireSalesforce, line 35: Does this query work? Usually we always use:
SELECT Id, PackageBundleId, VersionName, MajorVersion, MinorVersion, PackageBundleVersionId

Not the PackageBundle.Id format.

src/interfaces/bundleSObjects.ts
```diff
@@ -23,19 +24,54 @@ export namespace BundleSObjects {
...
+      Id: string;
+      BundleName: string;
+    };
+    PackageBundleVersion: {
+      Id: string;
+    };
+    VersionName: string;
+    MajorVersion: string;
+    MinorVersion: string;
+    'Ancestor.Id': string;
```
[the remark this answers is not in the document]
dawson-david-salesforce: Changed it so it is consistent, with the others. Additionally made it the type BundleVersion.

src/package/packageBundleVersionCreate.ts
dawson-david-salesforce: The Query does work, and I now rely on it slightly because it allows for me to have types of Bundle and BundleVersion within my requests, you can see my meaning better in the commit after this one. . Though if I did know that initially I would of gone with your way. Thank you for telling me.

btrn11 (approved):

PR #757, opened 2025-06-30
T/packaging distribution/w 18909971/package bundle delete

dawson-david-salesforce:
allow for deleteing bundles


src/package/packageBundle.ts
```diff
@@ -59,6 +59,26 @@ export class PackageBundle {
     return PackageBundleVersion.create(options, polling);
   }
 
+  public static async delete(connection: Connection, project: SfProject, idOrAlias: string): Promise<SaveResult> {
```
btrn11, line 62: let's add test

src/package/packageBundle.ts
```diff
@@ -59,6 +59,26 @@ export class PackageBundle {
...
+      throw new Error('Project instance is required when deleting package bundle by alias');
+    }
+
+    // Otherwise, treat it as an alias and resolve it from sfdx-project.json
+    const bundleId = project.getPackageBundleIdFromAlias(idOrAlias);
+    if (!bundleId) {
+      throw new Error(`No package bundle found with alias: ${idOrAlias}`);
+    }
+
+    return connection.tooling.sobject('PackageBundle').delete(bundleId);
```
btrn11, line 79: do we need to return here or the method can be void

test/package/resources/bundleProject/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
btrn11, line 1: do we need to commit these 2 json files? Can you confirm with the CLI team?

dawson-david-salesforce: Tested it, and we do not, I managed to use the packaingProject instead. Though if I implement more nut tests in the future I might bring it back. The main thing that we needed was a working sfdx.json.

btrn11 (approved):

PR #762, opened 2025-07-22
W-17561980 Outputs DOT Code for DisplayDependencies Command

MimiRaiSalesforce:
[@W-17561980@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000027hh2GYAQ/view)

# Summary
- Created nodes and edges according to the DependencyGraphJson. Nodes contain additional information including Package Name, Version Number, 04t ID. Edges contain a 04t source string and 04t target string.
- Call the produce() function to output nodes and edges in DOT code format and included output according to verbose and edge direction flag


messages/package_version_dependency.md
```diff
@@ -0,0 +1,11 @@
+# invalidPackageVersionIdError
+
+Can't display package dependencies. The package version ID %s you specified is invalid. Review the package version ID and then retry this command.
+
+# transitiveDependenciesRequiredError
+
+Can't display package dependencies. To display package dependencies, you must first add the calculateTransitiveDependencies parameter to the sfdx-project.json file, and set the value to "true". Next, create a new package version and then run this command using the 04t ID for the new package version.
+
+# invalidDependencyGraphError
```
MimiRaiSalesforce, line 9: We originally decided on using this error message if the DependencyGraphJson field was null from the Package2VersionCreateRequest table. However, I decided to make this error more general to also occur if values in the json are invalid (ex. cannot find a 04t listed in the JSON and therefore cannot query correct package name, version, etc. information)

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+      }
+      const package2VersionId = result05i.records[0].Id;
+
+      // Finally resolve to the Package2VersionCreateRequest ID (08c)
+      const query08c = `SELECT Id FROM Package2VersionCreateRequest WHERE Package2VersionId = '${package2VersionId}'`;
+      const result08c = await this.connection.tooling.query<{ Id: string }>(query08c);
+      if (result08c.records?.length !== 1) {
+        throw messages.createError('invalidPackageVersionIdError', [userPackageId]);
+      }
+      this.resolvedPackageId = result08c.records[0].Id;
```
luop90, line 101: IMO both of these bits (87-93 and 96-101) should both be extracted into their own functions. 

A "Clean Code" paradigm I subscribe to that functions help with self-documenting code in addition to code reuse. Thus, if you ever find yourself doing:

```
// Comment describing what the code below does
<code>
```

you should refactor this into:
```
def wellNamedFunctionDescribingTheCode:
    <code>
```

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+
+  /**
+   * Checks that the given Package2VersionCreateRequest ID (08c)
+   * 1) exists for the given devhub org
+   * 2) contains the calculateTransitiveDependencies boolean set to true
+   * 3) has a corresponding DependencyGraphJson
+   *
+   * @returns true if DOT code can be generated, false otherwise.
+   */
+  private async validatePackageVersion(): Promise<boolean> {
```
luop90, line 115: Another good place to break up this function into smaller functions, maybe combining the last two into one since they're similar.

E.g., something like:
```
verifyCreateRequestExistsOnDevHub()
verifyTransitiveDependenciesCalculated()
```

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+        }
+      } else {
+        throw messages.createError('transitiveDependenciesRequiredError');
+      }
+    }
+    return false;
+  }
+}
+
+export class DependencyDotProducer {
```
luop90, line 147: Maybe split this class into its own file?

test/package/packageVersionDependencies.test.ts
```diff
@@ -0,0 +1,391 @@
...
+ */
+import { expect } from 'chai';
+import { MockTestOrgData, TestContext } from '@salesforce/core/testSetup';
+import { Connection, Messages } from '@salesforce/core';
+import { SfProject } from '@salesforce/core/project';
+import sinon from 'sinon';
+import { PackageVersionDependency, VERSION_BEING_BUILT } from '../../src/package/packageVersionDependency';
+
+Messages.importMessagesDirectory(__dirname);
+// const messages = Messages.loadMessages('@salesforce/packaging', 'package_version_dependency');
```
luop90, line 15: Comment left behind?

test/package/packageVersionDependencies.test.ts
```diff
@@ -0,0 +1,391 @@
...
+  });
+
+  it('should throw an error if transitive dependency json is null', async () => {
+    connectionStub.onFirstCall().resolves({
+      records: [
+        {
+          CalcTransitiveDependencies: true,
+          DependencyGraphJson: null,
+        },
+      ],
```
luop90, line 133: These lines of code are identical to the test below it, with the sole exception of:

```
      expect.fail('Expected TransitiveDependenciesRequiredError to be thrown');
```
 vs
```
      expect.fail('Expected InvalidDependencyGraphError to be thrown');
```

There should be some way to remove the duplicate code. Since there's only two cases, a function (w/ a parameter for the error message name) is probably easiest

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+    const nodePromises = jsonNodes.map(async (node) => this.createSingleDependencyGraphNode(node.id));
+    const resolvedNodes = await Promise.all(nodePromises);
+    return resolvedNodes;
+  }
+
+  /**
+   * Creates a single dependency graph node.
+   * All nodes in the json are 04t... or VERSION_BEING_BUILT and requires different queries to create the node
+   */
+  private async createSingleDependencyGraphNode(nodeId: string): Promise<DependencyGraphNode> {
```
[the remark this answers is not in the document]
MimiRaiSalesforce: I created the functions `createVersionBeingBuiltNode` and `createNormalDependencyNode`.

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+    );
+    await producer.init();
+    return producer;
+  }
+
+  /**
+   * Resolves id input to a 08c. User can input a 08c or 04t.
+   * Currently a 04t is not supported in filtering the Package2VersionCreateRequest. So a user's input of 04t will be resolved to a 05i and then a 08c.
+   */
+  private async resolvePackageId(): Promise<void> {
```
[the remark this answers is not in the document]
MimiRaiSalesforce: I renamed to `resolvePackageCreateRequestId()`
While at it, I also decided to rename the instance variables `packageId` and `resolvedPackageId` to `userPackageVersionId` and `resolvedPackageVersionId`

src/package/packageVersionDependency.ts
MimiRaiSalesforce: I separated the functions to `query05iFrom04t()` and `query08cFrom05i`.

src/package/packageVersionDependency.ts
MimiRaiSalesforce: I decided to leave the classes all together for the following reasons:
- I decided to follow a similar structure to the packageAncestry class which contained multiple classes in the same file.
- The majority of the files in this `package` directory directly correspond to an sf cli command and followed a `packageCommandName.ts` naming structure. The files that deviated from this structure (ex. `subscriberPackageVersion.ts`, `versionNumber.ts`) contained classes that were used in multiple other classes. Splitting the other classes in this file (eg. `DependencyDotProducer` and `PackageDependencyNode`) would add unnecessary files while only being used within the `PackageVersionDependency`. 

TLDR: `DependencyDotProducer` and `PackageDependencyNode` are used very closely and only with `PackageVersionDependency` and should remain together for cleanliness.

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+      const record = nodeResult.records[0];
+      if (
+        !record.Package2?.Name ||
+        record.MajorVersion === null ||
+        record.MinorVersion === null ||
+        record.PatchVersion === null ||
+        record.BuildNumber === null
+      ) {
+        throw messages.createError('invalidDependencyGraphError');
+      }
```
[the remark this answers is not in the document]
MimiRaiSalesforce: I created the `throwErrorOnInvalidRecord()` function.

src/package/packageVersionDependency.ts
```diff
@@ -0,0 +1,382 @@
...
+    }
+    return `\t node_${targetNodeId} -> node_${sourceNodeId}`;
+  }
+}
+
+/**
+ * A class that represents a package dependency node.
+ * Given a PackageRequestId (08c), PackageName, SubscriberPackageVersionId (04t), and version (MajorVersion.MinorVersion.PatchVersion.BuildNumber) it will build a node of the package dependency.
+ */
+export class PackageDependencyNode extends AsyncCreatable<DependencyGraphNode> {
```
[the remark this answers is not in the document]
MimiRaiSalesforce: Added some comments above1

test/package/packageVersionDependencies.test.ts
MimiRaiSalesforce: Deleted 😬

test/package/packageVersionDependencies.test.ts
MimiRaiSalesforce: The difference between the two test cases is also that the values connectionStub resolves `CalcTransitiveDependencies` and `DependencyGraphJson` to. While I code create an additional function and pass in the parameters of these values and the expected errors, I don't think it would simplify the tests cases that significantly if only 2 tests use this function.

test/package/packageVersionDependencies.test.ts
```diff
@@ -0,0 +1,391 @@
...
+        {
+          SubscriberPackageVersionId: '04tXXXXXXXXXXXXXX2',
+          Package2: { Name: 'DependencyPackage2' },
+          MajorVersion: 1,
+          MinorVersion: 2,
+          PatchVersion: 3,
+          BuildNumber: 0,
+        },
+      ],
+    });
```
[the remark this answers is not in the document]
MimiRaiSalesforce: This helped a lot! There were 3 main resolutions which I simplified to the functions: `resolveDependencyGraphJsonCall()`, `resolveVersionBeingBuiltNodeCall()`, and `resolveDependencyNodeCall()`

Due to the multiple checks to validate the 04t ID, calculateTransitiveDependencies, and DependencyGraphJson, `resolveDependencyGraphJsonCall()` is called multiple times. The best resolution I could think of to prevent writing the same call multiple times was to pass in an array of callIndices.

luop90 (approved):

PR #765, opened 2025-07-29
W-18963108 Add Color for Selected Nodes in DisplayDependencies

MimiRaiSalesforce:
[@W-18963108@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002HQRhkYAH/view)

# Summary
- Include color for selected nodes given that the user specified package version is created (aka there is a 04t for the package version)
- Updated unit tests to resolve for the selected nodes when coloring nodes

### DIFF
--- a/src/package/packageVersionDependency.ts
+++ b/src/package/packageVersionDependency.ts
@@ -171,6 +171,7 @@ export class DependencyDotProducer {
   private subscriberPackageVersionId: string;
   private connection: Connection;
   private dependencyGraphData!: DependencyGraphData;
+  private selectedNodeIds: string[] = [];
 
   public constructor(
     connection: Connection,
@@ -205,6 +206,8 @@ export class DependencyDotProducer {
       nodes: await this.createDependencyGraphNodes(dependencyGraphJson.nodes),
       edges: this.createDependencyGraphEdges(dependencyGraphJson.edges),
     };
+
+    this.selectedNodeIds = await this.addSelectedNodeIds();
   }
 
   public produce(): string {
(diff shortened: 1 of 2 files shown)


src/package/packageVersionDependency.ts
```diff
@@ -218,6 +221,37 @@ export class DependencyDotProducer {
     return `strict digraph G {${EOL}${dotLines.join(EOL)}${EOL}}`;
   }
 
+  private async addSelectedNodeIds(): Promise<string[]> {
+    const selectedNodes: string[] = [];
+    if (this.subscriberPackageVersionId === VERSION_BEING_BUILT) {
```
MimiRaiSalesforce, line 226: If the package version isn't created (no 04t), only the VERSION_BEING_BUILT node is colored green

luop90 (approved):

src/package/packageVersionDependency.ts
```diff
@@ -218,6 +221,37 @@ export class DependencyDotProducer {
...
+          return selectedNodes;
+        }
+        if (dependencies.ids && Array.isArray(dependencies.ids)) {
+          const dependencyIds = dependencies.ids
+            .map((dep) => dep.subscriberPackageVersionId)
+            .filter((id) => id && typeof id === 'string' && id.startsWith('04t'));
+          selectedNodes.push(...dependencyIds);
+        }
+      } catch (error) {
+        throw messages.createError('invalidPackageVersionIdError', [this.subscriberPackageVersionId]);
```
shetzel, line 249: @MimiRaiSalesforce - throwing away the original error is non recommended.  You lose the entire stack trace and context.  The user is presented with just 1 possible cause (invalid package version ID) and not a good way to fix things (retry).  What if it was due to something else?  The error would then be wrong and misleading.

Also, this is a library and not necessarily executed by a CLI so library error messages should not refer to command execution.

PR #767, opened 2025-08-06
feat: added xgbRegression predictions for package push upgrades

dawson-david-salesforce:


src/package/packageUpgradePredict.ts
btrn11, line 1: @dawson-david-salesforce Why is this file so big?

dawson-david-salesforce: Because it is 3 ML models exported into typescript.

PR #870, opened 2026-05-27
@ W-22495755@ [CLI] Improve Error Message When Package Version Retrieve Fails Due to Dev Zip Not Generated

ravipanguluri:
This PR updates the error message that users receive when the developer zip failed to generate during conversion, and when the user tries to retrieve a dev zip for a native 2GP version for which there's no dev zip generated. Previously, the message was too long, and it misled users who saw dev zip generation failures during conversion because it did not tell users specifically to retry conversion. To make it clear, we split the long message into two distinct messages: one for converted 2GP package versions and one for native 2GP package versions. Each method includes specific, actionable remedies for the user. See [@W-22495755@](https://gus.lightning.force.com/lightning/_classic/%2Fa07EE00002a2ka3YAA)

### DIFF
--- a/messages/package.md
+++ b/messages/package.md
@@ -38,9 +38,13 @@ Can't retrieve package version metadata. The specified directory must be relativ
 
 Can't retrieve package metadata. To use this feature, you must first assign yourself the DownloadPackageVersionZips user permission. Then retry retrieving your package metadata.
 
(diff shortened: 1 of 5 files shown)


andrewslater (approved):
Looks good.

jshackell-sfdc (approved):
lgtm

PR #416, opened 2023-10-09
fix: add scopeProfiles feature to package version create

shetzel:
What does this PR do?
Adds support for `scopeProfiles` on package directories, which will only include profile metadata from the directory being packaged.

What issues does this PR fix or reference?
@W-14169751@

This PR depends on these PRs:
[fix: add scopeProfiles property to package directories](https://github.com/forcedotcom/schemas/pull/80)
[fix: add scopeProfiles to PackageDir type](https://github.com/forcedotcom/sfdx-core/pull/957)

PR #887, opened 2026-06-23
fix: scopeProfiles excludes profiles correctly with ./-prefixed paths @W-22149938@

WillieRuemmele:
## Summary
- **Bug 1**: `findAllProfiles` built glob ignore patterns that couldn't match `./`-prefixed directory paths from `sfdx-project.json` (e.g. `**/./src-access-management/**` never matches anything). Fixed by normalizing paths before building the glob pattern.
- **Bug 2**: `excludedProfiles` returned from `generateProfiles()` were full file path stems (e.g. `/tmp/.../profiles/Admin`) but the manifest filter compared them against profile member names (e.g. `Admin`). Fixed by extracting `path.basename()` before comparison.

Fixes https://github.com/forcedotcom/packaging/issues/416

## Work Item
@W-22149938@: Package version create with scopeProfiles: true includes profile names of other package directories in package.xml

## Proof of Work
- Tests: 41 passing (0 failing)
- Lint: 0 errors (pre-existing warnings only)
- Type check: clean
- New test added for ./-prefixed path exclusion

## Test plan
- [ ] Verify `sf package version create` succeeds with `scopeProfiles: true` when other package directories use `./`-prefixed paths
- [ ] Verify profiles from excluded directories do not appear in the generated `package.xml`
- [ ] Verify profiles that have no matching metadata nodes are correctly excluded from `package.xml` members

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   ],
   "dependencies": {
     "@jsforce/jsforce-node": "^3.10.16",
-    "@salesforce/core": "^8.31.1",
+    "@salesforce/core": "^8.31.2",
     "@salesforce/kit": "^3.2.6",
     "@salesforce/schemas": "^1.10.3",
     "@salesforce/source-deploy-retrieve": "^12.36.2",
--- a/src/package/packageProfileApi.ts
+++ b/src/package/packageProfileApi.ts
@@ -123,7 +123,10 @@ export class PackageProfileApi extends AsyncCreatable<ProfileApiOptions> {
 
   // Look for profiles in all package directories
   private findAllProfiles(excludedDirectories: string[] = []): string[] {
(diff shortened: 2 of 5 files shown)


soridalac (approved):

soridalac:
**QA notes**:
Using `scopeProfileIssue ` folder provided:

✅: Package Version created successfully 
`sf package version create --package bkpk-components --installation-key-bypass --wait 30 --target-dev-hub my-hub-org --verbose`
Screenshot 2026-06-26 at 2 11 01 PM
✅: Package Installed, no errors
✅: Profiles correctly excluded from bkpk component package.
✅: Configuration setting preserved
- ✅ scopeProfiles: true (still set)
- ✅ default: true (still set)
- ✅ versionDescription: "backpack design system component library" (preserved)

PR #889, opened 2026-07-02
fix: disambiguate package version retrieve errors (@W-23089623@)

ravipanguluri:
## What & Why

`sf package version retrieve` returned a single misleading error —
"you must first assign yourself the DownloadPackageVersionZips user permission" —
for three genuinely different failure modes:

1. The 04t ID doesn't exist anywhere.
2. The package version exists, but not in the targeted Dev Hub.
3. The user genuinely lacks the `DownloadPackageVersionZips` permission.

Only case 3 is a permission problem, so cases 1 and 2 sent users down a
fruitless perm-debugging detour. This splits the conflated check so each
case reports an actionable message.

@W-23089623@
Work item: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002cd3UWYAY/view

## How

The retrieve flow in `packageVersionRetrieve.ts` now runs two separate
`Package2Version` queries instead of one conflated lookup:

- **Disambiguation query** (`resolvePackage2Version`) selects only columns any
  authenticated user can read (`Package2Id`, `ConvertedFromVersionId`), so it
  never throws before we can classify the failure. If no row is visible in the
  running user's Dev Hub, it probes the global `SubscriberPackageVersion` view
  (queryable by any authenticated user) to tell **not-found**
  (`packageVersionNotFound`) apart from **exists-but-not-in-this-Dev-Hub**
  (`packageVersionNotInDevHub`).
- **Download-URL query** (`fetchDeveloperUsePkgZipUrl`) fetches the
  permission-controlled `DeveloperUsePkgZip` in an isolated query, run only
  after the row is confirmed present. Because existence is already
  established, both a "No such column" error and an empty value reliably mean
  the user lacks the permission → `developerUsePkgZipFieldUnavailable`
  (unchanged).

Selecting `DeveloperUsePkgZip` up front (as before) makes the query throw
"No such column" for a user without the permission, which is exactly what
masked cases 1 and 2 as a permission problem. Keeping it out of the
disambiguation query is what restores the three distinct messages.

Two new message keys added to `messages/package.md`:
`packageVersionNotFound`, `packageVersionNotInDevHub`.

## Tests

Updated/added unit coverage in `packageVersionMetadataRetrieve.test.ts`:
not-found, not-in-this-Dev-Hub, and missing-permission (empty value and
"No such column").

### DIFF
--- a/messages/package.md
+++ b/messages/package.md
@@ -38,6 +38,14 @@ Can't retrieve package version metadata. The specified directory must be relativ
 
 Can't retrieve package metadata. To use this feature, you must first assign yourself the DownloadPackageVersionZips user permission. Then retry retrieving your package metadata.
 
+# packageVersionNotFound
+
+Can't retrieve package metadata. We can't find the package version %s. Verify that the 04t ID is correct and that the package version exists.
+
+# packageVersionNotInDevHub
+
+Can't retrieve package metadata. Package version %s isn't accessible from this Dev Hub org. You can only retrieve package metadata from the Dev Hub that created the package version. Verify that you specified the correct target Dev Hub.
+
 # downloadDeveloperPackageZipHasNoDataNative2GP
 
 Can't retrieve package metadata. The developer package zip for this native 2GP package version is unretrievable. To resolve, create a new package version with the --generate-pkg-zip flag. Then retry retrieving your package metadata.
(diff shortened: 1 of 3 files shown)


jshackell-sfdc (approved):
lgtm (messages only!)

iowillhoit (approved):

PR #916, opened 2026-08-25
feat: adds limited package support for authorization cmds @W-23970417@

tshopshireSalesforce:


src/package/packageAuthorization.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+  return trimTo15(subscriberOrg);
+};
+
+export class PackageAuthorization {
+  private readonly connection: Connection;
+  private readonly subscriberPackageId?: string;
+
+  public constructor(options: PackageAuthorizationOptions) {
+    if (options.connection.getApiVersion() < MINIMUM_API_VERSION) {
```
btrn11: `getApiVersion()` returns a string, so this is a lexicographic comparison. Once the API reaches 100.0, `"100.0" < "68.0"` evaluates to `true` and blocks all authorization operations. Please compare numerically, e.g. `Number(options.connection.getApiVersion()) < Number(MINIMUM_API_VERSION)`.

src/package/packageAuthorization.ts
```diff
@@ -0,0 +1,132 @@
...
+      throw messages.createError('apiVersionTooLow', [MINIMUM_API_VERSION]);
+    }
+    if (
+      options.subscriberPackageId &&
+      (!options.subscriberPackageId.startsWith('033') || !validateSalesforceId(options.subscriberPackageId))
```
btrn11, line 54: `validateSalesforceId` in the pinned `@salesforce/core@9.0.0` is not anchored: `/[a-zA-Z0-9]{18}|[a-zA-Z0-9]{15}/`. It accepts malformed 18-character inputs such as `033000000000001!@@`, which are then interpolated into Tooling SOQL. Please add an anchored validation here, such as `/^(?:[A-Za-z0-9]{15}|[A-Za-z0-9]{18})$/`, before storing or querying with the value.

src/package/packageAuthorization.ts
sshyamala-sfdc: Just curious why do we want to do app version check? if app version is less then entity won't be visible.

src/package/packageAuthorization.ts
tshopshireSalesforce: I'd rather rely on the core library validation instead of adding our own so the behavior is consistent.

src/package/packageAuthorization.ts
tshopshireSalesforce: Will update.

tshopshireSalesforce: To provide a friendlier error message, it isn't necessary.

btrn11 (approved):

PR #920, opened 2026-08-27
feat: add PackageTrustLink.status for Public Secure link state (W-23970571)

sshyamala-sfdc:
## What & Why

Implements the library side of `sf package trust link status` (**W-23970571**, SPI / Public Secure).

Adds a read-only `PackageTrustLink.status()` that reports the connected authoring org's Public Secure (VerifiedDev) trust-link state so the developer can see where a link stands in the two-way handshake. Trust and enforcement remain decoupled — this never mutates anything.

## Behavior

Queries the authoring org's single `PkgVrfyAuthOrgTrustRela` record (the Tooling API query is scoped to the connected org, so it isn't filtered by verified org):

- **No record** → `{ Status: 'Not Linked', linked: false }`.
- **Record present** → its `Status` (`Pending`, `Accepted`, `Declined`, `Revoked`, or `Failed`), `linked: true`, the link Id, verified org ID, and only the timestamps that are set: `RequestedDate` (record `CreatedDate`), `EstablishedDate`, `RevokedDate`. Unset timestamps are omitted so `--json` stays clean.

## Files

- `src/package/packageTrustLink.ts` — new `status()` method; shared query extended with `EstablishedDate`, `RevokedDate`, `CreatedDate`.
- `src/interfaces/packagingInterfacesAndType.ts` — new `PackageTrustLinkStatusResult` type.
- `test/package/packageTrustLink.test.ts` — 5 unit tests.

## Testing

`yarn build` (compile + lint) clean; `PackageTrustLink` unit suite **10 passing** (Not Linked, full timestamps, Pending timestamp-omit, Revoked, query-not-scoped-by-verified-org).

## Related

Sibling verbs of `sf package trust link`: `request` (merged), `unlink` (#919). Companion plugin PR (the `sf package trust link status` command) pins the prerelease published from this branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@salesforce/packaging",
-  "version": "5.0.9-spi.0",
+  "version": "5.0.9-spi.1",
   "description": "Packaging library for the Salesforce packaging platform",
   "main": "lib/exported",
   "types": "lib/exported.d.ts",
--- a/src/interfaces/packagingInterfacesAndType.ts
+++ b/src/interfaces/packagingInterfacesAndType.ts
@@ -225,6 +225,24 @@ export type PackageTrustLinkRequestResult = {
   Status: string;
 };
 
+export type PackageTrustLinkStatusResult = {
+  // Current state of the authoring org's trust link: 'Not Linked' when no link exists, otherwise one
+  // of the trust-link statuses (Pending, Accepted, Declined, Revoked, Failed).
+  Status: string;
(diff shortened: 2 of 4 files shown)


btrn11 (approved):

sshyamala-sfdc:
GitHub release failed because of a duplicate naming issue. The commit auto-incremented the build to x.2, which is already taken by another pull request. However, this is not a problem because x.1 already has all the packaging changes required for the plugin module.

PR #928, opened 2026-09-04
@W-24017820@ fix(package-convert): actionable error when Dev Hub lacks 2GP

agayakwad-salesforce:
@W-24017820@

## What does this PR do?

Makes `sf package convert` (1GP → 2GP conversion) fail with a clear, actionable
error when run against a Dev Hub org that does not have second-generation managed
packaging enabled.

Previously, the very first server call in the convert flow — the `Package2`
tooling query in `findOrCreatePackage2` — had no error handling and did not route
through the error-massaging utilities. When the Dev Hub lacked 2GP, the raw
tooling-API error propagated straight to the user:

```
INVALID_TYPE: sObject type 'Package2' is not supported. If you are attempting to
use a custom object, be sure to append the '__c' after the entity name. Please
reference your WSDL or the describe call for the appropriate names.
```

There was no guidance telling the user the real problem: 2GP packaging isn't
enabled on their Dev Hub.

## What's the fix?

In `src/package/packageConvert.ts`:

- **Primary:** Wrap the `Package2` tooling query in `findOrCreatePackage2` in a
  `try/catch`. When the error indicates `Package2` is not supported, throw the
  new, actionable `convertPackagingNotEnabledOnOrg` message instead of the raw
  error. Any other error is rethrown unchanged.
- **Secondary:** Guard the `Package2` create path — if the create result reports
  the same "not supported" condition, throw the actionable message rather than
  the generic `combineSaveErrors` output.
- Add a small shared helper, `isPackage2NotSupportedError`, that detects the
  condition from either an `Error` or a jsforce `SaveError` object. It matches on
  a **substring** (`sObject type 'Package2' is not supported.`) because the full
  server message may append custom-object WSDL boilerplate — consistent with the
  existing handling in `packageVersionRetrieve.ts`.

New user-facing message (`messages/package_version_create.md`):

> Can't convert package. The org you specified doesn't have the required
> second-generation packaging permission enabled. Enable this permission on your
> Dev Hub org, and try again.

This mirrors the existing `packagingNotEnabledOnOrg` message used by
`sf package version retrieve`, keeping behavior consistent across commands.

## Before / After

**Before**
```
Error (INVALID_TYPE):
SELECT Id, Name FROM Package2 WHERE ConvertedFromPackageId
                     ^
ERROR at Row:1:Column:22
sObject type 'Package2' is not supported. ...
```

**After**
```
Error (ConvertPackagingNotEnabledOnOrgError): Can't convert package. The org you
specified doesn't have the required second-generation packaging permission
enabled. Enable this permission on your Dev Hub org, and try again.
```

The raw `INVALID_TYPE` / "is not supported" text no longer leaks to the user
(verified in both human-readable and `--json` output).

## Scope

This change is intentionally limited to the **convert** flow. The native
`sf package version create` path is out of scope for this PR and is unchanged.

## Testing

- Added unit tests in `test/package/packageConvert.test.ts`:
  - `findOrCreatePackage2` throws the actionable error when the `Package2` query
    fails with "not supported".
  - `findOrCreatePackage2` throws the actionable error when the `Package2`
    **create** reports "not supported".
  - `convertPackage` surfaces the actionable error (end-to-end propagation).
  - Each asserts the exact actionable message and that the raw
    `is not supported` text does not leak.
- Unit tests: `packageConvert.test.ts` — 25 passing.
- Lint and Prettier: clean on all changed files.
- **Manual end-to-end verification** against a locally-provisioned Dev Hub with
  2GP disabled: confirmed the convert command returns
  `ConvertPackagingNotEnabledOnOrgError` with no raw `INVALID_TYPE` leak in both
  default and `--json` output.

## Files changed

- `src/package/packageConvert.ts`
- `messages/package_version_create.md`
- `test/package/packageConvert.test.ts`

## Issues

@W-24017820@

### DIFF
--- a/messages/package_version_create.md
+++ b/messages/package_version_create.md
@@ -115,6 +115,10 @@ No subscriber package was found for seed id: %s
 Only one package in a Dev Hub is allowed per converted from first-generation package, but the following were found:
 %s
 
+# convertPackagingNotEnabledOnOrg
+
+Can't convert package. The org you specified doesn't have the required second-generation packaging permission enabled. Enable this permission on your Dev Hub org, and try again.
+
 # errorMissingPackageIdOrPath
 
 You must specify either a package ID or a package path to create a new package version.
--- a/src/package/packageConvert.ts
+++ b/src/package/packageConvert.ts
@@ -65,7 +65,15 @@ export async function findOrCreatePackage2(
   project?: SfProject
 ): Promise<string> {
   const query = `SELECT Id, Name FROM Package2 WHERE ConvertedFromPackageId = '${seedPackage}'`;
-  const queryResult = (await connection.tooling.query<PackagingSObjects.Package2>(query)).records;
+  let queryResult;
+  try {
+    queryResult = (await connection.tooling.query<PackagingSObjects.Package2>(query)).records;
+  } catch (e) {
+    if (isPackage2NotSupportedError(e)) {
+      throw messages.createError('convertPackagingNotEnabledOnOrg');
+    }
+    throw e;
+  }
   if (queryResult?.length > 1) {
     const ids = queryResult.map((r) => r.Id);
     throw messages.createError('errorMoreThanOnePackage2WithSeed', [ids.join(', ')]);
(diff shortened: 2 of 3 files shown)


agayakwad-salesforce:
> This PR lacks any commits of the 'fix' or 'feat' type, and therefore will not trigger a release. To silence all further warnings, react to this warning comment (or any other) with the 👀 emoji.
> 
> NOTE: If your repo uses squash commits, make sure to add the appropriate conventional commit prefix (e.g., 'fix:', 'feat:', 'chore:', etc) in the squash commit message.

resolved

